PR #23, opened 2021-01-19
ISSUE-8 Always activate glowroot

chibenwa:
- [x] TODO make it optional

Drafted before leaving office - to be continued tomorrow.


openpaas-james/apps/memory/pom.xml
```diff
@@ -112,6 +157,11 @@
                                 <from>src/main/conf</from>
                                 <into>/root/conf</into>
                             </path>
+                            <path>
+
```
Arsnael, line 161: extra line

Arsnael:
>     * [ ]  TODO make it optional

Do we really need to make it optional? Sure it's better but I would guess also that we want glowroot in general in all our deployments right?

chibenwa:
I can try to invest ~1h on it IMO

chibenwa:
```
docker run -d -p 4000:4000 -e "JAVA_TOOL_OPTIONS=-javaagent:/root/glowroot/glowroot.jar" linagora/openpaas-james-memory:latest
```

Screenshot from 2021-01-20 12-25-15

```
docker run -d -p 4000:4000 linagora/openpaas-james-memory:latest
```

Screenshot from 2021-01-20 12-24-20

openpaas-james/apps/memory/pom.xml
```diff
@@ -96,13 +140,14 @@
...
                             <port>587</port> <!-- SMTPS -->
-                            <port>4000</port> <!-- GLOWROOT, if activated -->
+                            <port>4000</port> <!-- GLOWROOT -->
                             <port>8000</port> <!-- WEBADMIN -->
                         </ports>
                         <appRoot>/root</appRoot>
                         <jvmFlags>
                             <jvmFlag>-Dlogback.configurationFile=/root/conf/logback.xml</jvmFlag>
                             <jvmFlag>-Dworking.directory=/root/</jvmFlag>
+                            <jvmFlag>-javaagent:/root/glowroot/glowroot.jar</jvmFlag>
```
Arsnael, line 150: I don't get how it is disabled by default with this in the JIB declaration?

remk (approved):

openpaas-james/apps/memory/pom.xml
chibenwa: See the wall diff @Arsnael :-)

chibenwa: I messed up and did not remove the line in the end (I mean: did not commit it)

Arsnael (approved):

chibenwa:
Rebased

tanandy:
are we sure there is no impact on performance ?

tanandy:
do we keep the possibility to enable it with env var ? we may want to disable it on production

rouazana:
> do we keep the possibility to enable it with env var ? we may want to disable it on production

It has been made optional in the second commit: https://github.com/linagora/openpaas-james/pull/23/commits/2743b8c9b379418d43a5016b0603d5bdef8d3090

Arsnael:
The title is misleading yes. I had impact on performance and yes we likely don't want it on prod, but as @rouazana said it has been made optional so no worries :)

PR #25, opened 2021-01-21
ISSUE-9 Memory deployment tests

chibenwa:
Result of today's MOB

Depends on https://github.com/linagora/james-project/pull/4222

### DIFF
--- a/openpaas-james/deployment-tests/common/src/main/java/com/linagora/openpaas/deployment/CliContract.java
+++ b/openpaas-james/deployment-tests/common/src/main/java/com/linagora/openpaas/deployment/CliContract.java
@@ -1,0 +1,33 @@
+package com.linagora.openpaas.deployment;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.Container;
+import org.testcontainers.containers.GenericContainer;
+
+public interface CliContract {
+    GenericContainer<?> jamesContainer();
+
+    @Test
+    default void cliShouldWork() throws Exception {
+        Container.ExecResult exec1 = jamesContainer().execInContainer("james-cli", "ListDomains");
+        SoftAssertions.assertSoftly(softly -> {
+            softly.assertThat(exec1.getExitCode()).isEqualTo(0);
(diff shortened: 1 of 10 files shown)


openpaas-james/deployment-tests/common/src/main/java/com/linagora/openpaas/deployment/CliContract.java
```diff
@@ -0,0 +1,18 @@
...
+import org.testcontainers.containers.GenericContainer;
+
+public interface CliContract {
+    GenericContainer<?> jamesContainer();
+
+    @Test
+    default void cliShouldWork() throws Exception {
+        jamesContainer().execInContainer("james-cli", "addDomain", "domain.tld");
+
+        assertThat(jamesContainer().execInContainer("james-cli", "ListDomains").getStdout())
```
rouazana, line 15: I'm not sure testing only stdout gives me enough confidence in this test. Shouldn't we test we really added the domain?

remk (approved):

openpaas-james/deployment-tests/common/src/main/java/com/linagora/openpaas/deployment/CliContract.java
Arsnael: how? by checking if webadmin works, which should maybe be an other test?

chibenwa: I would, **please**, like to keep the deployment test counts **as low as possible** as they are really long to play.

I am unwilling to have more than a single test for the CLI (which by the way is a net gain!)

Maybe that means breaking other testing rules like single assertion.

Let me suggest:

```
    default void cliShouldWork() throws Exception {
        assertThat(jamesContainer().execInContainer("james-cli", "ListDomains").getStdout())
            .doesNotContain("domain.tld");
            
        jamesContainer().execInContainer("james-cli", "addDomain", "domain.tld");

        assertThat(jamesContainer().execInContainer("james-cli", "ListDomains").getStdout())
```

Doing asserts on return codes is an option too.

remk (approved):

rouazana (approved):

chibenwa:
Squashed

PR #44, opened 2021-02-17
Reorganize stages in Jenkinsfile

Arsnael:
It made no sense to have the submodule fetch and the compilation of the project in the test stage of the pipeline.

It will increase readability in our CI pipeline UI for the job.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,7 @@ pipeline {
         stage('Test') {
             steps {
                 dir("openpaas-james") {
-                    sh 'mvn -B test'
+                    sh 'mvn -B surefire:test'
                 }
             }
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,13 +5,15 @@ pipeline {
         stage('Git checkout') {
             steps {
                 git 'https://github.com/linagora/openpaas-james'
+                sh 'git submodule init'
+                sh 'git submodule update'
             }
+        }
+        stage('Compile') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -5,13 +5,15 @@ pipeline {
...
+        stage('Compile') {
+            sh 'mvn install -Dmaven.javadoc.skip=true -DskipTests -T1C'
+        }
         stage('Test') {
             steps {
-                sh 'git submodule init'
-                sh 'git submodule update'
-                sh 'mvn install -Dmaven.javadoc.skip=true -DskipTests -T1C'
                 dir("openpaas-james") {
                     sh 'mvn -B test'
```
chibenwa, line 18: test -> surefire:test ?

Arsnael: what is the subtility here?

chibenwa: test runs validate stage, and do some extra work while surefire:test just plays tests

Arsnael: Nice thanks for the tip

chibenwa (approved):

aduprat (approved):

Arsnael:
> Checkstyle is run in the mvn clean install so no, it do not, and we do not need it.

Make me think... Should I add a clean in the mvn install current command? As maven is in the slave, and probably better to clean correctly previously generated packages?

chibenwa (approved):

Arsnael:
Merged

PR #58, opened 2021-03-08
Docker push master

Arsnael:
Gonna break for now likely... I didn't update the CI and this PR is not the master branch. I will add additional commits for testing with time


Arsnael:
Working. For my test I did kick out the when branch master directive and commented the tests, but I believe it will work now, but needs to be merged on master.

Help review thanks

LanKhuat (approved):

PR #71, opened 2021-03-12
WIP: ISSUE-47 DistributedServer should not depend on james-server-guice-d…

LanKhuat:
…istributed module.

Continuation of: https://github.com/linagora/openpaas-james/pull/66

Rechecking the code i found out that james-server-guice-distributed module has james-server-cassandra-guice as a dependency so I removed it and added stuffs that were missing.

### DIFF
--- a/openpaas-james/apps/distributed/pom.xml
+++ b/openpaas-james/apps/distributed/pom.xml
@@ -79,12 +79,16 @@
         <dependency>
             <groupId>${james.groupId}</groupId>
             <artifactId>blob-s3-guice</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>blob-s3-guice</artifactId>
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
-            <artifactId>data-cassandra</artifactId>
+            <artifactId>event-bus-distributed</artifactId>
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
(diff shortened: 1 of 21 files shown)


openpaas-james/apps/distributed/pom.xml
```diff
@@ -119,10 +123,6 @@
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-guice-data-ldap</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.james</groupId>
-            <artifactId>james-server-guice-distributed</artifactId>
```
chibenwa, line 124: No this one is perfectly fine: it is not a product but Guice modules for James as a toolkit...

openpaas-james/apps/distributed/pom.xml
```diff
@@ -108,7 +112,7 @@
             <artifactId>james-server-mailbox-plugin-deleted-messages-vault-guice</artifactId>
         </dependency>
         <dependency>
-            <groupId>${james.groupId}</groupId>
```
Arsnael, line 111: keep that :)

openpaas-james/apps/distributed/src/main/java/com/linagora/openpaas/james/app/BlobStoreCacheModulesChooser.java
```diff
@@ -0,0 +1,83 @@
+package com.linagora.openpaas.james.app;
```
Arsnael, line 1: Isn't that file copied from apache project? You need to keep the apache license on top then

Idem for similar files

chibenwa:
It is depending on guice-cassandra - that's fine (only modules)

It is not cassandra-guice (product)

openpaas-james/pom.xml
```diff
@@ -196,7 +201,7 @@
                 <version>${james.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.apache.james</groupId>
```
Arsnael, line 199: keep that

LanKhuat:
We should close this since it's no longer needed?

PR #81, opened 2021-03-19
Add missing logback json dependency in distributed version

Arsnael:
We need it to generate logs in json for fluentbit to be able to scrap them


chibenwa:
We have it for memory right?

chibenwa (approved):

Arsnael:
> We have it for memory right?

Thanks for the reminder, I did forget that one. Fixed now :)

LanKhuat (approved):

PR #100, opened 2021-03-31
ISSUE-98 POC PGP encryption

chibenwa:
Aaaaaand!

```
interview1@interview1-HP-ProBook-440-G6:~/vpn$ gpg --decrypt ttt
gpg: encrypted with 3072-bit RSA key, ID BB0AC3FE9D05E2C1, created 2021-03-31
      "Benoit Tellier <btellier@linagora.com>"
ga bou zo meuh
Bienvenue chez les shadocks.
interview1@interview1-HP-ProBook-440-G6:~/vpn$ 
```

Enjoy ;-)


Arsnael (approved):

chibenwa:
We imo need to move this into a folder `openpaas-james/mailbox/encrypted/pgp-utils`

Arsnael:
We should choose, PGP or GPG? we have a bit of mix here it seems:
```
openpaas-james/mailbox/encrypted/pgp-utils/src/main/java/com/linagora/openpaas/gpg/Encrypter.java
```
Module `pgp-utils` but in the `gpg` folder

openpaas-james/mailbox/encrypted/pgp-utils/src/test/java/com/linagora/openpaas/gpg/EncrypterTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public static final String INPUT = "ga bou zo meuh\r\nBienvenue chez les shadocks.\r\n";
+
+    @BeforeAll
+    static void setUpAll() throws Exception {
+        String bouncyCastleProviderClassName = "org.bouncycastle.jce.provider.BouncyCastleProvider";
+        Security.addProvider((Provider)Class.forName(bouncyCastleProviderClassName).getDeclaredConstructor().newInstance());
+    }
+
+    @Test
+    void encryptEmptyDataShouldBeSupported() throws Exception {
```
LanKhuat, line 26: What happen if i use a wrong public/private key pair? Or a key that does not conform to PGP format?

chibenwa: > What happen if i use a wrong public/private key pair?

The email can't be decrypted.

> Or a key that does not conform to PGP format?

Read key would throw

I'll add the tests

LanKhuat (approved):

Arsnael (approved):

PR #113, opened 2021-04-14
ISSUE-114 Encrypt tests fix

Arsnael:


LanKhuat (approved):

chibenwa (changes requested):
https://github.com/linagora/openpaas-james/blob/master/openpaas-james/mailbox/encrypted/api/pom.xml

```
        <dependency>
            <artifactId>james-server-testing</artifactId>
        </dependency>
```

should be in test scope!

Arsnael:
> https://github.com/linagora/openpaas-james/blob/master/openpaas-james/mailbox/encrypted/api/pom.xml
> 
> ```
>         <dependency>
>             <groupId>${james.groupId}</groupId>
>             <artifactId>james-server-testing</artifactId>
>         </dependency>
> ```
> 
> should be in test scope!

And this is what you get  when you compile with this dep in test scope:

```
java: package org.junit.jupiter.api does not exist
```

So no...

chibenwa (approved):

Arsnael:
About the james-server-testing scope thing... 

There is a much more general pattern present in the project after checking quickly... As the scope `test` is not transitive, it seems usually we have this dep introduced without the scope test in the common module of integration and deployment tests, to not have to repeat this dep in other modules testing the implementation of those contracts, in james project and tmail also. 

Might need a separated change. Shouldn't the junit deps using the scope test as well while we are thinking about it? it's not the case for now it seems in james-server-testing pom module.

chibenwa:
>  it seems usually we have this dep introduced without the scope test in the common module of integration and deployment tests, to not have to repeat this dep in other modules testing the implementation of those contracts, in james project and tmail also. Is this intended?

Yes it is intended, are only tests depends from integration contract mvn modules. No polution of released artifacts, and no boiler plate.

It is not the case here...

> Might need a separated change. Shouldn't the junit deps using the scope test as well while we are thinking about it? it's not the case for now it seems in james-server-testing pom module (and maybe others)

james-server-testing is only used in scope test OR by mvn modules only relied on by tests, no big deal.

BTW I forgot about this: can we get a ticket number here before a merge?

Arsnael:
> BTW I forgot about this: can we get a ticket number here before a merge?

https://github.com/linagora/openpaas-james/issues/114

vttranlina (approved):

PR #115, opened 2021-04-15
ISSUE-92 Keystore Cassandra implementation

LanKhuat:


openpaas-james/mailbox/encrypted/api/src/main/scala/com/linagora/openpaas/encrypted/KeystoreManager.scala
```diff
@@ -10,7 +10,27 @@ object KeyId {
 
 case class KeyId(value: String)
 
-case class PublicKey(id: KeyId, payload: Array[Byte])
+class PublicKey(id: KeyId, payload: Array[Byte]) {
```
Arsnael, line 13: Why not keeping the case class here? I don't see how is this commit useful^^'

LanKhuat: the different is the equals method

chibenwa: > the different is the equals method

Bad answer. I prefer a separate "have same content method"

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/CassandraKeystoreDAO.scala
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.james.backends.cassandra.utils.CassandraAsyncExecutor
+import org.apache.james.core.Username
+import reactor.core.scala.publisher.{SFlux, SMono}
+
+class CassandraKeystoreDAO @Inject()(executor: CassandraAsyncExecutor,
+                                     insertStatement: PreparedStatement,
+                                     selectStatement: PreparedStatement,
+                                     selectAllStatement: PreparedStatement,
+                                     deleteStatement: PreparedStatement,
+                                     deleteAllStatement: PreparedStatement) {
```
chibenwa, line 21: Please don't pass statements in the constructor, you will not be able to initialize them with an inject.

Do directly vals in the class. CF CassandraTaskExecutionDetailsProjectionDAO

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/CassandraKeystoreDAO.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  def this(session: Session, cassandraTypesProvider: CassandraTypesProvider) {
```
chibenwa, line 23: Kill that constructor

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/CassandraKeystoreDAO.scala
```diff
@@ -0,0 +1,74 @@
...
+      executor = new CassandraAsyncExecutor(session),
+      insertStatement = session.prepare(insertInto(TABLE_NAME)
+        .value(USERNAME, bindMarker(USERNAME))
+        .value(ID, bindMarker(ID))
+        .value(KEY, bindMarker(KEY))),
+
+      selectStatement = session.prepare(select.from(TABLE_NAME)
+        .where(QueryBuilder.eq(USERNAME, bindMarker(USERNAME)))
+        .and(QueryBuilder.eq(ID, bindMarker(ID)))
+        .allowFiltering),
```
chibenwa, line 34: `allowFiltering` 

Euhhhh are you sure????

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/CassandraKeystoreDAO.scala
```diff
@@ -0,0 +1,74 @@
...
+
+      selectAllStatement = session.prepare(select.from(TABLE_NAME)
+        .where(QueryBuilder.eq(USERNAME, bindMarker(USERNAME)))
+        .orderBy(asc(ID))),
```
chibenwa, line 38: We definitly do not need ordering!

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/table/CassandraKeystoreModule.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.datastax.driver.core.schemabuilder.SchemaBuilder.rows
+import com.linagora.openpaas.encrypted.cassandra.table.KeyStoreTable.{ID, KEY, TABLE_NAME, USERNAME}
+import org.apache.james.backends.cassandra.components.CassandraModule
+import org.apache.james.backends.cassandra.utils.CassandraConstants.DEFAULT_CACHED_ROW_PER_PARTITION
+
+object CassandraKeystoreModule {
+  val MODULE: CassandraModule = CassandraModule.table(TABLE_NAME)
+    .comment("Holds Keystore definition. Used to manage user public keys used for data encryption.")
+    .options((options: Options) => options
+      .clusteringOrder(ID, ASC)
```
chibenwa, line 17: We do not need order

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/CassandraKeystoreDAO.scala
LanKhuat: Cassandra will throw without allowFiltering option. USERNAME is PRIMARY KEY and the query is filtering on ID field, no? May be we need to define a COMPOSITE KEY (USERNAME + ID)?

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/CassandraKeystoreDAO.scala
LanKhuat: then why define ID as CLUSTERING COLUMN?

chibenwa: To be able to list IDs per user

openpaas-james/mailbox/encrypted/cassandra/src/main/java/com/linagora/openpaas/encrypted/cassandra/CassandraKeystoreDAO.scala
chibenwa: ID being a clustering column you should not have this behaviour. If you have it then you are doing something wrong.

openpaas-james/mailbox/encrypted/api/src/main/scala/com/linagora/openpaas/encrypted/KeystoreManager.scala
Arsnael: For the array bytes? Scala can't compare two array bytes out of the box? (maybe it can't I don't know^^)

Arsnael (approved):

openpaas-james/mailbox/encrypted/api/src/main/scala/com/linagora/openpaas/encrypted/KeystoreManager.scala
```diff
@@ -5,22 +5,25 @@ import org.apache.james.core.Username
...
-    def fromPayload(payload: Array[Byte]): KeyId = KeyId(BaseEncoding.base16().encode(payload))
+  def fromPayload(payload: Array[Byte]): KeyId = KeyId(BaseEncoding.base16().encode(payload))
 }
 
 case class KeyId(value: String)
 
-case class PublicKey(id: KeyId, payload: Array[Byte])
+case class PublicKey(id: KeyId, payload: Array[Byte]) {
+  def hasSameContentAs(publicKey: PublicKey): Boolean =
+    id.equals(publicKey.id) && new String(payload).equals(new String(publicKey.payload))
```
chibenwa, line 15: Use Array.equals?

openpaas-james/mailbox/encrypted/cassandra/src/test/java/CassandraKeystoreManagerTest.java
```diff
@@ -0,0 +1,36 @@
...
+import com.linagora.openpaas.encrypted.cassandra.CassandraKeystoreManager;
+import com.linagora.openpaas.encrypted.cassandra.table.CassandraKeystoreModule;
+
+public class CassandraKeystoreManagerTest implements KeystoreManagerContract {
+
+    @RegisterExtension
+    static CassandraClusterExtension cassandraCluster = new CassandraClusterExtension(
+        CassandraModule.aggregateModules(CassandraKeystoreModule.MODULE(),
+            CassandraSchemaVersionModule.MODULE,
+            CassandraZonedDateTimeModule.MODULE));
```
chibenwa, line 21: We likely do not need CassandraZonedDateTimeModule.MODULE

chibenwa (approved):

PR #126, opened 2021-04-19
ISSUE-95 Common contract for MessageManagerTest with InMemory/Cassandra keystore

LanKhuat:


chibenwa:
I don't see the value added of this "integration tests at the mailbox level" - all subparts are already tested, only the key mechanisms storage differs.

If I'm picky I would say the only benefit here is to run tests a second time, on a slower technology (Cassandra).

I would prefer to write "end to end integration tests at the protocol level" to be honest.

LanKhuat:
After discussion, we agreed that this change does not bring much value to production code.

PR #130, opened 2021-04-22
ISSUE-96 Injection for the EncryptedMailbox

chibenwa:
https://github.com/linagora/openpaas-james/pull/127 without the integration test


chibenwa:
Abort. Base PR is :green_apple:  with reuse forks.

PR #140, opened 2021-05-04
ISSUE-99 Buffer encrypted stream before writing it on the file system

chibenwa:
This prevents frequent disk flushes harming performance.

Needs to be backed by performance tests before being adopted...


chibenwa:
`chibenwa/openpaas-james-distributed:gpg-buffer` can be used to test the performance profile of this cc @Arsnael

Arsnael (approved):
Ok will try to test that this morning

Arsnael:
Tests proved that we get closer performances between no encryption and encryption with this change.

Results here: https://github.com/linagora/openpaas-james/issues/101#issuecomment-832409729

PR #155, opened 2021-05-28
[PERFORMANCE] Rely on MIME4J 0.8.5-SNAPSHOT for TMail

chibenwa:
The newest commits shipped on MIME4J ships some nice-to-have performance improvments we might want to benefit from 
prior a mime4J release that anyway takes time.

In the meantime, TMAIL, not subject to the release pluging not accepting `-SNAPSHOT` (a bocker for `apache/james-project`) can start using the fully compatible version of MIME4J.


chibenwa:
Failing until snapshots gets published.

Arsnael (approved):

PR #159, opened 2021-05-30
ISSUE-158 Manage James dependencies with a BOM

chibenwa:


tmail-backend/pom.xml
```diff
@@ -236,647 +195,49 @@
...
             </dependency>
             <dependency>
                 <groupId>${james.groupId}</groupId>
-                <artifactId>james-server-guice-data-ldap</artifactId>
-                <version>${james.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>org.apache.james</groupId>
-                <artifactId>james-server-guice-distributed</artifactId>
+                <artifactId>jmap-rfc-8621-integration-tests-common</artifactId>
```
Arsnael, line 220: Do we still need this then?

Arsnael (commented):
Can you explain a bit more how it works then? I'm confused at to what to review exactly

tmail-backend/pom.xml
```diff
@@ -236,647 +195,49 @@
...
-                <artifactId>james-server-guice-configuration</artifactId>
-                <version>${james.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>${james.groupId}</groupId>
-                <artifactId>james-server-guice-common</artifactId>
+                <artifactId>james-project</artifactId>
                 <version>${james.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
```
chibenwa, line 209: Please maven do take the deoendency management section of the pom.xml of `james-project` and put it here as if I had been copy and pasting it.

tmail-backend/pom.xml
chibenwa: Not defined in the BOM POM thus we need an explicit statement.

tmail-backend/pom.xml
chibenwa: #72laBalle

Arsnael: Why need to import the BOM of james-server-guice then?I think all submodules of james-server-guice are being declared in the james-project maven file no?

tmail-backend/pom.xml
```diff
@@ -236,647 +195,49 @@
...
             </dependency>
             <dependency>
                 <groupId>${james.groupId}</groupId>
-                <artifactId>james-server-guice-common</artifactId>
-                <type>test-jar</type>
+                <artifactId>james-server-guice</artifactId>
```
Arsnael, line 213: Why need to import the BOM of james-server-guice then?I think all submodules of james-server-guice are being declared in the james-project maven file no?

chibenwa: THat is what I did.

chibenwa: Screenshot from 2021-05-31 10-57-24

Arsnael: you did not really answered my question but ok I checked a bit myself more in details and indeed there is extra modules declared in james-server-guice...

Arsnael (approved):

PR #162, opened 2021-06-02
Update James project SHA1

Arsnael:


tmail-backend/mailbox/encrypted/api/src/main/scala/com/linagora/tmail/encrypted/EncryptedMessageManager.scala
```diff
@@ -81,4 +67,22 @@ class EncryptedMessageManager @Inject()(messageManager: MessageManager, keystore
     messageManager.getMetaData(resetRecent, mailboxSession, fetchGroup)
 
   override def getResolvedAcl(mailboxSession: MailboxSession): MailboxACL = messageManager.getResolvedAcl(mailboxSession)
+
+  private def append(appendCommand: AppendCommand, session: MailboxSession): SMono[AppendResult] =
+    SFlux.fromPublisher(keystoreManager.listPublicKeys(session.getUser))
+      .collectSeq()
+      .map(keys => {
+        if (keys.isEmpty) {
+          messageManager.appendMessage(appendCommand, session)
```
chibenwa, line 76: flatMap and use the appendMessageReactive please

tmail-backend/mailbox/encrypted/api/src/main/scala/com/linagora/tmail/encrypted/MailboxReactorUtils.java
```diff
@@ -0,0 +1,57 @@
...
+package com.linagora.tmail.encrypted;
+
+import java.util.Optional;
+
+import org.apache.james.mailbox.exception.MailboxException;
+import org.reactivestreams.Publisher;
+
+import reactor.core.publisher.Mono;
+
+public abstract class MailboxReactorUtils {
```
chibenwa, line 29: Why can't we reuse `org.apache.james.mailbox.store.MailboxReactorUtils` ?

Am I missing something?

chibenwa: Right.

PR #193, opened 2021-07-19
Update James project SHA1

Arsnael:
And from there I will be able to tag the release


Arsnael:
I'm on the issue, should be easy to fix :)

chibenwa (approved):

PR #202, opened 2021-07-29
[UPGRADE] Switch to MIME4J 0.8.5

chibenwa:


tmail-backend/pom.xml
```diff
@@ -41,7 +41,6 @@
         <james.groupId>org.apache.james</james.groupId>
         <james.version>3.7.0-SNAPSHOT</james.version>
         <james.protocols.groupId>${james.groupId}.protocols</james.protocols.groupId>
-        <apache-mime4j.version>0.8.5-SNAPSHOT</apache-mime4j.version>
```
Arsnael, line 44: I would think it's more of a cleanup? That declared version was not used anywhere if I'm not wrong, thus the version used was the one declared by apache james project?

Arsnael (approved):

tmail-backend/pom.xml
chibenwa: No it override the versions used elsewhere.

Arsnael: ah ok my bad then :)

vttranlina (approved):

PR #203, opened 2021-07-30
James upgrade 14

Arsnael:


tmail-backend/blob/blobid-list/src/test/java/com/linagora/tmail/blob/blobid/list/SingleSaveBlobStoreTest.java
```diff
@@ -56,4 +69,22 @@ public BlobIdList blobIdList() {
...
         return BucketName.DEFAULT;
     }
+
+    @Override
+    @Test
+    public void listBucketsShouldReturnAllBucketsInUse() {
+        BlobStoreDAO store = testee();
+
+        Mono.from(store.save(TEST_BUCKET_NAME, TEST_BLOB_ID, SHORT_BYTEARRAY)).block();
+        Mono.from(store.save(CUSTOM_BUCKET_NAME, OTHER_TEST_BLOB_ID, SHORT_BYTEARRAY)).block();
```
Arsnael, line 79: Had to override this test, because in the original we use `TEST_BLOB_ID`, but the `SingleSaveBlobStore` sees that we already have a blob with the same id (created line before) and skips it. 

However, it means the single save blob store does not take buckets into account. My question is: is this normal? I would feel like we should take the bucket into account and that there could be a same blob in different buckets perhaps? Or maybe I forgot the reason for this case...

chibenwa: If I remember correctly single save should be on only on the default bucket and do nothing for non-default buckets.

Can you check that this is the actual behaviour?

Arsnael: That's not the impression I have at all from the code. We just pass the bucketName as a param without checking anything to the underlying S3BlobStore... 

This test proves as well that we create two buckets in which we save different blobs. And we return the list with 2 buckets. I don't think we do nothing for non-default buckets

chibenwa: Crap. Then we might need a bugfix...

Arsnael: Can we merge this though meanwhile and maybe create a ticket to fix it?

chibenwa: Create ticket first!

Arsnael: https://github.com/linagora/tmail-backend/issues/204

chibenwa (approved):

Arsnael:
Squashed fixup

PR #271, opened 2021-09-30
ISSUE-208 Accept keypair as private key

chibenwa:


tungtv202 (approved):

Arsnael (approved):

Arsnael:
I agree with the logic, but shouldn't we test it?

chibenwa:
> I agree with the logic, but shouldn't we test it?

Yes, feel free to contribute tests.

chibenwa (commented):
:+1:

PR #274, opened 2021-10-04
WIP ISSUE-260 Implement TeamMailboxUserEntityValidator

Arsnael:
I push just because we say we push at the end of the day... but not much to see as:
* I didn't realize at first but I need this to be done: https://github.com/linagora/tmail-backend/issues/254
* there is an other PR coming changing https://github.com/apache/james-project/pull/678
* we need to wait at the end that the works on james-project are being merged to do a synced upgrade on the submodule in tmail

So of course a lot of things are missing :)


tmail-backend/mailbox/team-mailboxes/src/main/scala/com/linagora/tmail/team/TeamMailboxUserEntityValidator.scala
```diff
@@ -0,0 +1,36 @@
...
+import java.util
+import javax.inject.Inject
+
+object TeamMailboxUserEntityValidator {
+  val TEAM_MAILBOX: EntityType = new EntityType("team-mailbox")
+
+  val TEAM_MAILBOX_NAMESPACE: String = MailboxConstants.NAMESPACE_PREFIX_CHAR + "TeamMailbox"
+}
+
+class TeamMailboxUserEntityValidator @Inject()(mailboxManager: MailboxManager) extends UserEntityValidator {
```
chibenwa, line 20: You should use the API written by Tung...

tmail-backend/mailbox/team-mailboxes/src/main/scala/com/linagora/tmail/team/TeamMailboxUserEntityValidator.scala
```diff
@@ -0,0 +1,36 @@
...
+  override def canCreate(username: Username, ignoredTypes: util.Set[EntityType]): Option[UserEntityValidator.ValidationFailure] = {
+    if (ignoredTypes.contains(TEAM_MAILBOX)) {
+      return None
```
chibenwa, line 23: `return` spotted in Scala code

tmail-backend/mailbox/team-mailboxes/src/main/scala/com/linagora/tmail/team/TeamMailboxUserEntityValidator.scala
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    if (username.getDomainPart.isPresent) {
+      val mailboxSession: MailboxSession = mailboxManager.createSystemSession(username)
+      val checkPath: MailboxPath = new MailboxPath(TEAM_MAILBOX_NAMESPACE, Username.of(username.getDomainPart.get.asString()), username.getLocalPart + "." + MailboxConstants.INBOX)
```
chibenwa, line 28: `Username.of(username.getDomainPart.get.asString())`

What if the username do not have a domain part?

tmail-backend/mailbox/team-mailboxes/src/main/scala/com/linagora/tmail/team/TeamMailboxUserEntityValidator.scala
Arsnael: I know... API that is not available yet... just missed out that part when started writing yesterday

tmail-backend/mailbox/team-mailboxes/src/main/scala/com/linagora/tmail/team/TeamMailboxUserEntityValidator.scala
Arsnael: the purpose of `if (username.getDomainPart.isPresent) {` two lines above?

tmail-backend/mailbox/team-mailboxes/src/main/scala/com/linagora/tmail/team/TeamMailboxUserEntityValidator.scala
Arsnael: it's not a wrong syntax with conditions though... Scala does not know you need to return here in this case if you don't use return, as the code continues after the conditional.

If you mean instead be more functional, so I don't have to see a return here, then just say it :)

tmail-backend/mailbox/team-mailboxes/src/main/scala/com/linagora/tmail/team/TeamMailboxUserEntityValidator.scala
chibenwa: IMO in scala we should make the effort of writing fully functional code.

The main pain toward it in Java is unchecked exceptions. There's not this in Scala.

Arsnael: Agreed I know, but I realized I missed the all team mailbox api and that it was useless to put more efforts into this for now... But point taken

Arsnael:
Just missing a few test cases

chibenwa:
+ update ApacheJames submodule ;-)

Arsnael:
> No guice bindings for `TeamMailboxUserEntityValidator` ?

Was surprised no bindings for the routes, no module... so was wondering if bindings were done later or not?

PR #318, opened 2021-12-03
Update James project SHA1

Arsnael:


Arsnael:
```
[ERROR] shouldUseMemoryMailboxManager{GuiceJamesServer}  Time elapsed: 0.15 s  <<< ERROR!

com.github.fge.lambdas.ThrownByLambdaException: org.apache.commons.configuration2.ex.ConfigurationException: Unable to init mailet org.apache.james.jmap.mailet.VacationMailet

Caused by: org.apache.commons.configuration2.ex.ConfigurationException: Unable to init mailet org.apache.james.jmap.mailet.VacationMailet

Caused by: javax.mail.MessagingException: Can not load mailet org.apache.james.jmap.mailet.VacationMailet

Caused by: java.lang.ClassNotFoundException: org.apache.james.jmap.mailet.VacationMailet
```
ahah yeah forgot there was some refactoring on the vacation mailet as well... will fix this

chibenwa (approved):

chibenwa (approved):

tmail-backend/combined-identity/src/main/java/com/linagora/tmail/combined/identity/CombinedUsersRepositoryModule.java
```diff
@@ -42,6 +43,12 @@ public LdapRepositoryConfiguration provideConfiguration(ConfigurationProvider co
             configurationProvider.getConfiguration("usersrepository"));
     }
 
+    @Provides
+    @Singleton
+    public CassandraRepositoryConfiguration provideCassandraRepositoryConfiguration() {
+        return CassandraRepositoryConfiguration.DEFAULT;
```
Arsnael, line 49: There was a missing Guice binding regarding the `CombinedUsersRepository`. Now with the password hash salting refactoring we need a binding to `CassandraRepositoryConfiguration` as our CombinedUsersRepository is using CassandraUsersDAO. But we don't really care about password hashing configuration in this case, as we just want to store the username, we never really use the password there. The test for login still occurs against ldap.

So I binded a dummy default one. Is that ok @chibenwa ?

chibenwa: Looks great to me.

Arsnael:
Squashed

chibenwa (approved):

PR #329, opened 2022-01-18
ISSUE-316 Multiple fixes regarding contact search auto-complete

chibenwa:


chibenwa:
Cc @tieugiang91

tmail-backend/jmap/extensions-es/src/main/java/com/linagora/tmail/james/jmap/EmailAddressContactMappingFactory.java
```diff
@@ -0,0 +1,54 @@
...
+                    .endObject()
+                    .startObject("id")
+                        .field("type", "text")
+                    .endObject()
+                    .startObject("address")
+                        .field("type", "text")
+                    .endObject()
+                .endObject()
+            /*
+                // TODO add the analisis / tokenizer parameter as parameters of IndexCreationFactory...
```
Arsnael: Is that supposed to follow an update on James submodule?

chibenwa: No it requires a ticket and a contribution in /backends-common/es-v7

Arsnael: yeah that's what I had in mind with my question

PR #340, opened 2022-02-15
Add rate limiter dependency into Tmail

Arsnael:


Arsnael:
It works, as I could check locally. However, I was not sure if it was the best way or if we can do something smarter with maven here..?

tmail-backend/apps/memory/pom.xml
```diff
@@ -187,6 +187,39 @@
...
+                        <id>copy-rate-limiting-jar</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${basedir}/target/extensions-jars</outputDirectory>
+                            <resources>
+                                <resource>
+                                    <directory>../../../james-project/server/mailet/rate-limiter-redis/target</directory>
```
chibenwa, line 200: Let's use the in memory one only, no?

chibenwa (approved):
Ok if it works though adding a regular maven depency would have had worked too.

We could also register the rate limiters guice mudule directly in the guice bindings of the APP, so configuration is eased.

Arsnael:
> Ok if it works though adding a regular maven depency would have had worked too.
> 
> We could also register the rate limiters guice mudule directly in the guice bindings of the APP, so configuration is eased.

Hmmm I think I misunderstood. It might be easier and makes more sense indeed. Clearly looked for complicated for nothing here

Arsnael:
Needs https://github.com/apache/james-project/pull/883

quantranhong1999 (approved):

chibenwa (approved):

vttranlina (approved):

PR #355, opened 2022-03-14
Revert "ISSUE-361: Add trigger for Gatling-imap build"

Arsnael:
This is not working and causes the master build to fail, so I revert it for now


Arsnael:
Do not merge yet...

PR #384, opened 2022-03-30
EmailAddress autocompletion should handle domain contacts & should support index contact name

Arsnael:


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -1,24 +1,29 @@
...
+    EmailAddressContact(computeId(address), address, firstname, surname)
 }
 
-case class EmailAddressContact(id: UUID, address: MailAddress) {
-  def contains(part: String): Boolean = address.asString().contains(part)
+case class EmailAddressContact(id: UUID, address: MailAddress, firstname: String = "", surname: String = "") {
+  def contains(part: String): Boolean =
+    address.asString().contains(part) ||
+      (firstname != null && firstname.contains(part)) ||
+      (surname != null && surname.contains(part))
```
Arsnael, line 26: Maybe there is better?

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -29,21 +34,40 @@ case class AccountEmailContact(accountId: String, id: UUID, address: MailAddress
 trait EmailAddressContactSearchEngine {
   def index(accountId: AccountId, address: MailAddress): Publisher[EmailAddressContact]
 
+  def index(accountId: AccountId, address: MailAddress, firstname: String, surname: String): Publisher[EmailAddressContact]
+
+  def index(domain: Domain, address: MailAddress, firstname: String, surname: String): Publisher[EmailAddressContact]
```
Arsnael, line 39: I thought it was making sense to keep a personal index contact without names as I think it can be the case (we only type a mail address when composing a mail without a name associated to it)

However, I thought for domain likely not, as we fetch that from the ldap. But maybe there is cases when there is no name declared in a ldap? If you think yes, I could add back an extra method for domain and no name indexing

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -29,21 +34,40 @@ case class AccountEmailContact(accountId: String, id: UUID, address: MailAddress
...
-  override def autoComplete(accountId: AccountId, part: String): Publisher[EmailAddressContact] =
+  override def index(domain: Domain, address: MailAddress, firstname: String, surname: String): Publisher[EmailAddressContact] =
+    index(domain, EmailAddressContact.of(address, firstname, surname))
+
+  private def index(domain: Domain, addressContact: EmailAddressContact): Publisher[EmailAddressContact] =
+    SMono.fromCallable(() => domainList.put(domain, addressContact))
+      .`then`(SMono.just(addressContact))
+
+  override def autoComplete(accountId: AccountId, part: String): Publisher[EmailAddressContact] = {
+    val domain = Username.of(accountId.getIdentifier).getDomainPart.get // TODO: evaluate optional
```
Arsnael, line 67: Will work on it, the mix Scala, reactor, optional was a pain :)

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
chibenwa: IMO we should get rid of the null checks.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -29,21 +34,40 @@ case class AccountEmailContact(accountId: String, id: UUID, address: MailAddress
 trait EmailAddressContactSearchEngine {
   def index(accountId: AccountId, address: MailAddress): Publisher[EmailAddressContact]
 
+  def index(accountId: AccountId, address: MailAddress, firstname: String, surname: String): Publisher[EmailAddressContact]
```
chibenwa, line 37: This method has too much parameters, we likely should define  helper case class:

```
case class ContactFields(address: MailAddress, firstname: String, surname: String)
```

So that this then becomes:

```
def index(accountId: AccountId, fields: ContactFields): Publisher[EmailAddressContact]
```

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
chibenwa: IMO uneeded.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
chibenwa: Yes here we loose the ability to do JMAP without virtual hosting...

Let me suggest:

```
    val maybeDomain: Option[Domain] = Username.of(accountId.getIdentifier).getDomainPart.asScala // needs import OptionalUtils for scala-jdk8 compat

    SFlux.concat(
          maybeDomain..map(domain => SFlux.fromIterable(domainList.get(domain).asScala).getOrElse(SFlux.empty()),
          SFlux.fromIterable(emailList.get(accountId).asScala))
      .filter(_.contains(part))
```

quantranhong1999 (commented):
Read it. Nothing to add.

chibenwa (approved):
Looks overall good to me.

Please add another ticket to update, delete contacts.

tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
```diff
@@ -74,6 +70,6 @@ public ESEmailAddressContactSearchEngine(ReactorElasticSearchClient client, Elas
             .map(Arrays::asList)
             .flatMapIterable(Function.identity())
             .map(Throwing.function(hit -> new EmailAddressContact(UUID.fromString((String) hit.getSourceAsMap().get("id")),
-                new MailAddress((String) hit.getSourceAsMap().get("address")), "", "")));
+                new ContactFields(new MailAddress((String) hit.getSourceAsMap().get("address")), "", ""))));
```
quantranhong1999, line 73: I see we set the default value for firstname and surname. We can omit them at this moment?

Not a big deal though.

quantranhong1999 (approved):

Arsnael:
> Please add another ticket to update, delete contacts.

=> https://github.com/linagora/tmail-backend/issues/385

tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
Arsnael: Nope... I set the default values but it's in a scala class. This is Java though, you seem to need to declare them, the conversion does not work. If I omit them, I can't compile :)

PR #391, opened 2022-04-06
Update delete contact

Arsnael:
Will rebase it against Quan's PR tomorrow to have ES implem as well to take into account and autocomplete case ignore too, but let me know if the API is fine like this though, I wasn't completely sure

### DIFF
--- a/tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
+++ b/tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
@@ -7,16 +7,17 @@ import static com.linagora.tmail.james.jmap.ContactMappingFactory.EMAIL;
 import static com.linagora.tmail.james.jmap.ContactMappingFactory.FIRSTNAME;
 import static com.linagora.tmail.james.jmap.ContactMappingFactory.SURNAME;
 
+import java.util.List;
 import java.util.UUID;
 
 import javax.inject.Inject;
 import javax.mail.internet.AddressException;
 
-import org.apache.commons.lang3.NotImplementedException;
 import org.apache.james.backends.es.v7.DocumentId;
 import org.apache.james.backends.es.v7.ElasticSearchIndexer;
 import org.apache.james.backends.es.v7.ReactorElasticSearchClient;
 import org.apache.james.backends.es.v7.RoutingKey;
+import org.apache.james.backends.es.v7.UpdatedRepresentation;
 import org.apache.james.core.Domain;
 import org.apache.james.core.MailAddress;
 import org.apache.james.core.Username;
@@ -40,6 +41,7 @@ import com.linagora.tmail.james.jmap.dto.DomainContactDocument;
 import com.linagora.tmail.james.jmap.dto.UserContactDocument;
 
 import reactor.core.publisher.Mono;
+import reactor.util.function.Tuple2;
 
 public class ESEmailAddressContactSearchEngine implements EmailAddressContactSearchEngine {
     private static final String DELIMITER = ":";
@@ -79,22 +81,48 @@ public class ESEmailAddressContactSearchEngine implements EmailAddressContactSea
 
     @Override
     public Publisher<EmailAddressContact> update(AccountId accountId, MailAddress mailAddress, ContactFields updatedFields) {
-        throw new NotImplementedException("Not implemented yet!");
+        if (mailAddress.equals(updatedFields.address())) {
+            EmailAddressContact updatedEmailAddressContact = EmailAddressContact.of(updatedFields);
+            return Mono.fromCallable(() -> mapper.writeValueAsString(new UserContactDocument(accountId, updatedEmailAddressContact)))
+                .flatMap(content -> userContactIndexer.update(
+                    List.of(new UpdatedRepresentation(computeUserContactDocumentId(accountId, mailAddress), content)),
+                    RoutingKey.fromString(mailAddress.asString())))
+                .thenReturn(updatedEmailAddressContact);
+        } else {
+            return Mono.zip(Mono.from(delete(accountId, mailAddress)), Mono.from(index(accountId, updatedFields)))
+                .map(Tuple2::getT2);
+        }
     }
 
     @Override
     public Publisher<EmailAddressContact> update(Domain domain, MailAddress mailAddress, ContactFields updatedFields) {
-        throw new NotImplementedException("Not implemented yet!");
+        if (mailAddress.equals(updatedFields.address())) {
+            EmailAddressContact updatedEmailAddressContact = EmailAddressContact.of(updatedFields);
+            return Mono.fromCallable(() -> mapper.writeValueAsString(new DomainContactDocument(domain, updatedEmailAddressContact)))
+                .flatMap(content -> domainContactIndexer.update(
+                    List.of(new UpdatedRepresentation(computeDomainContactDocumentId(domain, mailAddress), content)),
+                    RoutingKey.fromString(mailAddress.asString())))
+                .thenReturn(updatedEmailAddressContact);
+        } else {
+            return Mono.zip(Mono.from(delete(domain, mailAddress)), Mono.from(index(domain, updatedFields)))
+                .map(Tuple2::getT2);
+        }
     }
 
     @Override
     public Publisher<Void> delete(AccountId accountId, MailAddress address) {
-        throw new NotImplementedException("Not implemented yet!");
+        return userContactIndexer.delete(
+                List.of(computeUserContactDocumentId(accountId, address)),
+                RoutingKey.fromString(address.asString()))
+            .then();
     }
 
     @Override
     public Publisher<Void> delete(Domain domain, MailAddress address) {
-        throw new NotImplementedException("Not implemented yet!");
+        return domainContactIndexer.delete(
+                List.of(computeDomainContactDocumentId(domain, address)),
+                RoutingKey.fromString(address.asString()))
+            .then();
     }
 
     @Override
--- a/tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
+++ b/tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
@@ -17,7 +17,6 @@ import org.apache.james.backends.es.v7.DocumentId;
 import org.apache.james.backends.es.v7.ElasticSearchIndexer;
 import org.apache.james.backends.es.v7.ReactorElasticSearchClient;
 import org.apache.james.backends.es.v7.RoutingKey;
-import org.apache.james.backends.es.v7.UpdatedRepresentation;
 import org.apache.james.core.Domain;
 import org.apache.james.core.MailAddress;
 import org.apache.james.core.Username;
@@ -41,7 +40,6 @@ import com.linagora.tmail.james.jmap.dto.DomainContactDocument;
 import com.linagora.tmail.james.jmap.dto.UserContactDocument;
 
 import reactor.core.publisher.Mono;
-import reactor.util.function.Tuple2;
 
 public class ESEmailAddressContactSearchEngine implements EmailAddressContactSearchEngine {
     private static final String DELIMITER = ":";
@@ -80,33 +78,13 @@ public class ESEmailAddressContactSearchEngine implements EmailAddressContactSea
     }
 
     @Override
-    public Publisher<EmailAddressContact> update(AccountId accountId, MailAddress mailAddress, ContactFields updatedFields) {
-        if (mailAddress.equals(updatedFields.address())) {
-            EmailAddressContact updatedEmailAddressContact = EmailAddressContact.of(updatedFields);
-            return Mono.fromCallable(() -> mapper.writeValueAsString(new UserContactDocument(accountId, updatedEmailAddressContact)))
-                .flatMap(content -> userContactIndexer.update(
-                    List.of(new UpdatedRepresentation(computeUserContactDocumentId(accountId, mailAddress), content)),
-                    RoutingKey.fromString(mailAddress.asString())))
-                .thenReturn(updatedEmailAddressContact);
-        } else {
-            return Mono.zip(Mono.from(delete(accountId, mailAddress)), Mono.from(index(accountId, updatedFields)))
-                .map(Tuple2::getT2);
-        }
+    public Publisher<EmailAddressContact> update(AccountId accountId, ContactFields updatedFields) {
+        return index(accountId, updatedFields);
     }
 
     @Override
-    public Publisher<EmailAddressContact> update(Domain domain, MailAddress mailAddress, ContactFields updatedFields) {
-        if (mailAddress.equals(updatedFields.address())) {
-            EmailAddressContact updatedEmailAddressContact = EmailAddressContact.of(updatedFields);
-            return Mono.fromCallable(() -> mapper.writeValueAsString(new DomainContactDocument(domain, updatedEmailAddressContact)))
-                .flatMap(content -> domainContactIndexer.update(
-                    List.of(new UpdatedRepresentation(computeDomainContactDocumentId(domain, mailAddress), content)),
-                    RoutingKey.fromString(mailAddress.asString())))
-                .thenReturn(updatedEmailAddressContact);
-        } else {
-            return Mono.zip(Mono.from(delete(domain, mailAddress)), Mono.from(index(domain, updatedFields)))
-                .map(Tuple2::getT2);
-        }
+    public Publisher<EmailAddressContact> update(Domain domain, ContactFields updatedFields) {
+        return index(domain, updatedFields);
     }
 
     @Override
--- a/tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
+++ b/tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
@@ -12,6 +12,7 @@ import java.util.UUID;
 import javax.inject.Inject;
 import javax.mail.internet.AddressException;
 
+import org.apache.commons.lang3.NotImplementedException;
 import org.apache.james.backends.es.v7.DocumentId;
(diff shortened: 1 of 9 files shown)


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -79,6 +82,26 @@ class InMemoryEmailAddressContactSearchEngine extends EmailAddressContactSearchE
     SMono.fromCallable(() => domainList.put(domain, addressContact))
       .`then`(SMono.just(addressContact))
 
+  override def delete(accountId: AccountId, mailAddress: MailAddress): Publisher[Void] =
+    getContact(accountId, mailAddress)
+      .map(contact => emailList.remove(accountId, contact))
+      .`then`()
+
+  private def getContact(accountId: AccountId, mailAddress: MailAddress): SMono[EmailAddressContact] =
```
quantranhong1999, line 102:
```suggestion
  private def getUserContact(accountId: AccountId, mailAddress: MailAddress): SMono[EmailAddressContact]
```

Or you can just overload this method ?

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -79,6 +82,26 @@ class InMemoryEmailAddressContactSearchEngine extends EmailAddressContactSearchE
...
+
+  override def delete(domain: Domain, mailAddress: MailAddress): Publisher[Void] =
+    getDomainContact(domain, mailAddress)
+      .map(contact => domainList.remove(domain, contact))
+      .`then`()
+
+  private def getDomainContact(domain: Domain, mailAddress: MailAddress): SMono[EmailAddressContact] =
+    SFlux.fromIterable(domainList.get(domain).asScala)
+      .filter(_.fields.address.equals(mailAddress))
+      .singleOrEmpty()
```
quantranhong1999, line 115: I think should we delete all the records (for idempotent).

FYI on the ES side, we have a unique ID for the document (domain + mail address) => reduce duplicated index data, just need to delete one record :') For memory implementation, I think we can use Guava Table, however not really strict though.

tmail-backend/jmap/extensions/src/test/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContactSearchEngineContract.scala
```diff
@@ -178,4 +178,58 @@ trait EmailAddressContactSearchEngineContract {
...
       .containsOnly(contactFieldsB, otherContactFields)
   }
+
+  @Test
+  def deleteShouldDeletePersonalContact(): Unit = {
+    SMono(testee().index(accountId, otherContactFields)).block()
+
+    SMono(testee().delete(accountId, otherMailAddress)).block()
+
+    assertThat(SFlux.fromPublisher(testee().autoComplete(accountId, "pen")).asJava().map(_.fields).collectList().block())
```
quantranhong1999, line 188:
```suggestion
    assertThat(SFlux.fromPublisher(testee().autoComplete(accountId, "Car")).asJava().map(_.fields).collectList().block())
```
A reminder here: for surname and firstname we use edge_ngram

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -82,6 +86,14 @@ class InMemoryEmailAddressContactSearchEngine extends EmailAddressContactSearchE
     SMono.fromCallable(() => domainList.put(domain, addressContact))
       .`then`(SMono.just(addressContact))
 
+  override def update(accountId: AccountId, mailAddress: MailAddress, updatedFields: ContactFields): Publisher[EmailAddressContact] =
+    SMono.fromPublisher(delete(accountId, mailAddress))
+      .`then`(SMono.fromPublisher(index(accountId, updatedFields)))
```
quantranhong1999, line 91: If we use Guava Table here, I guess we can optimize this by directly putting data instead of doing a delete before adding.

Arsnael: Do we really care of having a super optimized mem version though?

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
Arsnael: > I think should we delete all the records (for idempotent).

Hmmm why? I want to delete one contact for a domain, not all?

> For memory implementation, I think we can use Guava Table, however not really strict though.

No need I think

quantranhong1999: > Hmmm why? I want to delete one contact for a domain, not all?

Is there any case where we have many contacts with the same mail address in the real world?

Arsnael: Usually when you create a mail address at any mail provider, I believe if you try to choose a mail address that is already used by someone else, you get an error. For the good reason that if your server gets a mail addressed to bob@linagora.com but you have Bob and Alice linked to it, to whom do you deliver it? Mail addresses for a domain are unique I believe

quantranhong1999: > Mail addresses for a domain are supposed to be unique I believe

So that is why contact for mail address should be unique also. With current memory implementation, we could index many contacts with the same mail address => deleting just one contact is not enough.

This is what I mean :)

Arsnael: ah didn't get your point that way... makes sense :) not making the task easier but all right, I give you credit for this one ;)

quantranhong1999: Yeh, that is it. Let me handle it if you need some help :p 
For the memory implementation, you can have a look at `MemoryCustomIdentityDAO`.
I think we can create another ticket for implementing ES6/ES7 for these update/contact APIs if necessary?

Arsnael: As you wish, I'm banging my head at trying to merge latest james in tmail for now :D

chibenwa: Agree for indempotance.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
Arsnael: At the end... no. For the case that if the email address changes. It's better to just delete the old record and putting the new one. If we directly putting data it can create inconsistencies on the mail address if that one changes (and this case is valid)

Arsnael:
Will work on ES now

tmail-backend/jmap/extensions/pom.xml
```diff
@@ -19,6 +19,12 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>team-mailboxes</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>apache-james-backends-es-v7</artifactId>
```
Arsnael, line 24: Not found of that... However I need that to include QueryBuilder in the contract to test properly update with ES implem. I could not find any better... But I'm opened for ideas

tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
```diff
@@ -79,22 +79,48 @@ public ESEmailAddressContactSearchEngine(ReactorElasticSearchClient client, Elas
 
     @Override
     public Publisher<EmailAddressContact> update(AccountId accountId, MailAddress mailAddress, ContactFields updatedFields) {
-        throw new NotImplementedException("Not implemented yet!");
+        if (mailAddress.equals(updatedFields.address())) {
```
Arsnael, line 82: As the document ID is based on the mailAddress, if the mailAddress changes, we then need to delete and recreate a record. If only names change, we can do a normal update. Thus the if

tmail-backend/jmap/extensions-es-v7/src/main/java/com/linagora/tmail/james/jmap/ESEmailAddressContactSearchEngine.java
```diff
@@ -79,22 +79,48 @@ public ESEmailAddressContactSearchEngine(ReactorElasticSearchClient client, Elas
...
+            EmailAddressContact updatedEmailAddressContact = EmailAddressContact.of(updatedFields);
+            return Mono.fromCallable(() -> mapper.writeValueAsString(new UserContactDocument(accountId, updatedEmailAddressContact)))
+                .flatMap(content -> userContactIndexer.update(
+                    ImmutableList.of(new UpdatedRepresentation(computeUserContactDocumentId(accountId, mailAddress), content)),
+                    RoutingKey.fromString(mailAddress.asString())))
+                .thenReturn(updatedEmailAddressContact);
+        } else {
+            return Mono.from(delete(accountId, mailAddress))
+                .then(Mono.from(index(accountId, updatedFields)));
```
quantranhong1999, line 91: I think it would be more optimized if these two requests are done parallelly (the delete request not strictly be done first). Maybe something like this ?:
```java
            Mono<Void> deleteMono = Mono.from(delete(accountId, mailAddress));
            Mono<EmailAddressContact> indexMono = Mono.from(index(accountId, updatedFields));
            return Mono.zip(deleteMono, indexMono)
                .map(Tuple2::getT2);
```

rf: https://stackoverflow.com/questions/60850570/execute-three-mono-in-parallel-as-soon-they-are-created-wait-for-all-to-finish

An better solution maybe even just need to wait index be finished and don't care about the delete request finishment?

Arsnael: Hmmm I'm not sure that it's really necessary TBH... other opinions on this?

tmail-backend/jmap/extensions/pom.xml
vttranlina: we can add only elasticsearch client dependency?
```
        <dependency>
            <groupId>org.elasticsearch.client</groupId>
            <artifactId>elasticsearch-rest-high-level-client</artifactId>
            <version>7.10.2</version>
            <exclusions>
                <exclusion>
                    <groupId>commons-logging</groupId>
                    <artifactId>commons-logging</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

quantranhong1999: (y) I think this dependency is more lightweight.

tmail-backend/jmap/extensions/pom.xml
```diff
@@ -20,10 +20,16 @@
             <artifactId>team-mailboxes</artifactId>
         </dependency>
         <dependency>
-            <groupId>${james.groupId}</groupId>
-            <artifactId>apache-james-backends-es-v7</artifactId>
+            <groupId>org.elasticsearch.client</groupId>
+            <artifactId>elasticsearch-rest-high-level-client</artifactId>
+            <version>7.10.2</version>
```
vttranlina, line 25: Should we have a {property} or add version in parent pom?

quantranhong1999: For this one, I think it is no. Cause we have another version 6.4.3 for backport.

chibenwa:
Rebase needed

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
```diff
@@ -36,16 +35,19 @@ case class ContactFields(address: MailAddress, firstname: String = "", surname:
...
-}
-
 trait EmailAddressContactSearchEngine {
   def index(accountId: AccountId, fields: ContactFields): Publisher[EmailAddressContact]
 
   def index(domain: Domain, fields: ContactFields): Publisher[EmailAddressContact]
 
+  def update(accountId: AccountId, mailAddress: MailAddress, updatedFields: ContactFields): Publisher[EmailAddressContact]
+
+  def update(domain: Domain, mailAddress: MailAddress, updatedFields: ContactFields): Publisher[EmailAddressContact]
```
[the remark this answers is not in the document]
Arsnael: I think when talking about it with @chibenwa , we came to the conclusion that for example for well lsc sync with ldap if there is an email change, it would be translated by a delete and a create (your case). 

But then after I wasn't sure about what the third party app would send... If an update of mailaddress only was possible or not. So I did it in case.

But if we are sure that we only update the name, and in case the email change we are ok to delete and recreate the contact, then I'm ok to simplify this method, will make the logic easier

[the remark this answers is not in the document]
quantranhong1999: If we are not sure how the update works with third-party apps, maybe using Option for mailAddress is better for now?

>  if we are sure that we only update the name, and in case the email change we are ok to delete and recreate the contact, then I'm ok to simplify this method, will make the logic easier

Otherwise, I am OK with this.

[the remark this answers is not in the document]
Arsnael: Option would make sense in this case yes... not gonna simplify the code though :D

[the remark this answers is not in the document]
Arsnael: I tried something, let me know if it's ok for you!

quantranhong1999 (approved):

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContact.scala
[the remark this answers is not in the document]
Arsnael: Ok been agreed in grooming to simplify this and only reduce the update to name update.

If the mail address changes -> delete (old email address) + create (new email address)

vttranlina (approved):

Arsnael:
Team:

Do you think the last commit makes sense? I thought we want to index the document if we do an update on a non-existing entry (thus the test) but our ES client isn't configured for this. So I made a proposition to add an upsert argument to allow this behavior upon es updates. The default is false of course to not change the behavior for other indexes.

If it's ok, I will port the code back to James for ES7. WDYT?

tmail-backend/backend-commons/elasticsearch-v6/src/main/java/org/apache/james/backends/es/ElasticSearchIndexer.java
```diff
@@ -83,6 +83,10 @@ private void logContent(DocumentId id, String content) {
     }
 
     public Mono<BulkResponse> update(List<UpdatedRepresentation> updatedDocumentParts, RoutingKey routingKey) {
+        return update(updatedDocumentParts, routingKey, false);
+    }
+
+    public Mono<BulkResponse> update(List<UpdatedRepresentation> updatedDocumentParts, RoutingKey routingKey, boolean upsert) {
```
quantranhong1999, line 89: I remember that having a boolean in a method's arguments is a code smell that we should have 2 different methods instead (here upsert one and just update one).

quantranhong1999 (approved):
Me I am okay with the upsert. IMO we could integrate this change in a separate PR. (Cause the sprint is ending :p)

vttranlina:
> I thought we want to index the document if we do an update on a non-existing entry (thus the test) but our ES client isn't configured for this

yes, I agree

tmail-backend/backend-commons/elasticsearch-v6/src/main/java/org/apache/james/backends/es/ElasticSearchIndexer.java
Arsnael: I do have two methods?

quantranhong1999:
```
public Mono<BulkResponse> update(List<UpdatedRepresentation> updatedDocumentParts, RoutingKey routingKey)

public Mono<BulkResponse> upsert(List<UpdatedRepresentation> updatedDocumentParts, RoutingKey routingKey)
```

Maybe one private to avoid duplicate code.
```
private Mono<BulkResponse> update(List<UpdatedRepresentation> updatedDocumentParts, RoutingKey routingKey, boolean upsert)
```

Arsnael: Ah get it :) Good idea

Arsnael:
> Me I am okay with the upsert. IMO we could integrate this change in a separate PR. (Cause the sprint is ending :p)

That task wasn't part of the sprint so no big deal :)

tmail-backend/backend-commons/elasticsearch-v6/src/main/java/org/apache/james/backends/es/ElasticSearchIndexer.java
Arsnael: Done!

quantranhong1999 (approved):

Arsnael:
Upsert on James es indexer: https://github.com/apache/james-project/pull/965

Arsnael:
Forget the upsert, an index just does the same thing (documents are immutables in ES, it deletes the old one and recreates the new one anyways)

Hope it's the last change :)

Arsnael:
Squashed

chibenwa (approved):

PR #443, opened 2022-06-17
Use JDK-17 when compiling Apache James

chibenwa:


Jenkinsfile
```diff
@@ -18,7 +18,7 @@ pipeline {
         }
         stage('Compile') {
             steps {
-                sh 'mvn clean install -Dmaven.javadoc.skip=true -DskipTests -T1C'
+                sh 'mvn clean install -Dmaven.javadoc.skip=true -DskipTests -T1C -Dtarget.jdk=17'
```
Arsnael, line 21: as simple as that right? Nice :)

Arsnael (approved):

Jenkinsfile
chibenwa: Yes !

PR #453, opened 2022-08-02
ISSUE-441 Migrate from ES7 to Opensearch for TMail (high level rest client)

Arsnael:
WIP, I still  have issues, like quan, with some es6 testing modules...


tmail-backend/deployment-tests/common/src/main/java/com/linagora/tmail/deployment/ThirdPartyContainers.java
```diff
@@ -7,7 +7,7 @@
 
 public class ThirdPartyContainers {
     public static String ES6_IMAGE_NAME = "docker.elastic.co/elasticsearch/elasticsearch:6.3.2";
-    public static String ES7_IMAGE_NAME = "docker.elastic.co/elasticsearch/elasticsearch:7.10.2";
+    public static String OS_IMAGE_NAME = "opensearchproject/opensearch:2.0.0";
```
quantranhong1999, line 10:
```suggestion
    public static String OS_IMAGE_NAME = "opensearchproject/opensearch:2.1.0";
```

And other places.

Arsnael: Good catch!

quantranhong1999 (approved):

PR #470, opened 2022-09-06
[REFACTORING] Remove the binding on the Rspamd listener

Arsnael:
In case we are deploying TMail without rspamd, the listener will keep trying to connect to Rspamd for any mail activity which results in a lot of errors messages in logs.

It's still possible to add the listener via listeners.xml file


chibenwa (approved):

vttranlina (commented):
good job

PR #482, opened 2022-10-04
Update James SHA1

quantranhong1999:
for TMail 0.6.3 release


upgrade-instructions.md
```diff
@@ -14,4 +14,70 @@ Note: this section is in progress. It will be updated during all the development
...
\ No newline at end of file
+(No changes yet.)
+
+## 0.6.3 version
+
+Change list:
+
+- [Improve indexing of email address domains](#improve-indexing-of-email-address-domains)
+
+### Improve indexing of email address domains
```
Arsnael, line 25: I think that upgrade instruction should be part of apache james and not necessary here..?

quantranhong1999: We will need documentation upgrade instructions for `JAMES-3827 Backport to ElasticSearch 6.x` anyway.

Arsnael: Usually putting them on the changelog but... it's true that we don't use this upgrade-instructions in this project until now. Deserves a discussion :)

chibenwa: No not in james as we migrated from OpenSearch.

In TMail it is different as we have an intermediate OpenSearch release, which thus need an upgrade instruction.

chibenwa (approved):

Arsnael (approved):

Arsnael (approved):

vttranlina (approved):

PR #505, opened 2022-10-18
Update James-SHA1

Arsnael:


chibenwa:
The James SHA-1 you use is likely a failing build.

Maybe we should merge https://github.com/apache/james-project/pull/1260 first and re-update?

Arsnael:
> The James SHA-1 you use is likely a failing build.

It is not, as we don't retest James submodule.

PR #543, opened 2023-01-11
ISSUE-541 Autocomplete should not index/return duplicated contacts

quantranhong1999:


chibenwa (commented):
My feedback:
 - We do not need filtering at read time and write time. One or the other is sufficient.
 - I tend to prefer write time as read is performance critical and more often called. Having unbounded read requests on reads sucks IMO
  - If we want to get rids of duplicates at write time the most straight forward is likely to derive the DocumentId from the invariant (the address?) so that two contacts considered identical results in the same accountId. Then we can leverage idempotance and avoid read-before-writes.
   - https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html `op_type` `create` could even be used to avoid updating an existing document.

quantranhong1999:
> If we want to get rids of duplicates at write time the most straight forward is likely to derive the DocumentId from the invariant (the address?) so that two contacts considered identical results in the same accountId. Then we can leverage idempotance and avoid read-before-writes.

I am not sure if unique DocumentId applies for 2 indexes (we use `user` contact and `domain` contact), will see more tmr.

quantranhong1999:
I reworked a bit the work (mostly on read time deduplication).

> If we want to get rids of duplicates at write time the most straight forward is likely to derive the DocumentId from the invariant (the address?) so that two contacts considered identical results in the same accountId. 

We already do that DocumentId per index, but [ES does not support that unique DocumentId across indices](https://discuss.elastic.co/t/is-it-possible-to-make-document-id-unique-among-indices/117690) (we read from 2 indexes).
I could not find any better solution than read-before-writes for index time. We do it async so it is acceptable IMO.

> Having unbounded read requests on reads sucks IMO

I agree, staying away from unbounded read requests now.

> https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html op_type create could even be used to avoid updating an existing document.

It seems not applicable cause we do support updating contacts.

> We do not need filtering at read time and write time. One or the other is sufficient.

IMO we could leverage both:
- Deduplication at index time could save storage on ES while the impact on performance is not notable to users.
- Deduplication at read time could save some duplicated cases while the deduplication at index time can not avoid: 
  GIVEN user contact X is indexed WHEN we index domain contact X THEN we still need to index domain contact X cause not all users in the domain have contact X.
  Further, the performance impaction while doing deduplication at read time now is very small: we only further do distinct operations on a limited contact collection.
  The cons is with duplicated cases, we would not return exactly the `limit` contacts requested but return < `limit` - which is totally acceptable IMO.

chibenwa (approved):

chibenwa:
Needs a rebase.

Arsnael (approved):

PR #605, opened 2023-03-15
[FIX] Avoid blocking call by TMailLocalDelivery on Cassandra event loop thread

quantranhong1999:
- Cassandra's event loop was blocked 
```json
{"timestamp":"2023-03-14T08:18:02.828Z","level":"ERROR","thread":"s1-io-4","logger":"org.apache.james.transport.mailets.delivery.MailDispatcher","message":"Error while storing mail. This is a final exception.","context":"default","exception":"com.datastax.oss.driver.api.core.DriverTimeoutException: Query timed out after PT5S..."}
```
- Furthermore, MailDispatcher retries (if needed) delivering mail on Parallel scheduler, which would throw right away .block in the pipeline.

### DIFF
--- a/tmail-backend/mailets/src/main/scala/com/linagora/tmail/mailets/TMailMailboxAppender.scala
+++ b/tmail-backend/mailets/src/main/scala/com/linagora/tmail/mailets/TMailMailboxAppender.scala
@@ -1,15 +1,13 @@
 package com.linagora.tmail.mailets
 
 import com.linagora.tmail.team.{TeamMailbox, TeamMailboxRepository}
+import javax.mail.internet.MimeMessage
 import org.apache.james.core.Username
+import org.apache.james.mailbox.MailboxManager
 import org.apache.james.mailbox.MessageManager.AppendResult
 import org.apache.james.mailbox.model.ComposedMessageId
-import org.apache.james.mailbox.MailboxManager
 import org.apache.james.transport.mailets.delivery.MailboxAppenderImpl
 import reactor.core.publisher.Mono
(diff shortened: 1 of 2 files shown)


quantranhong1999:
TODO: run a performance test round to make sure `com.datastax.oss.driver.api.core.DriverTimeoutException` related to `MailDispatcher` is not existed.

chibenwa (approved):

chibenwa:
```
[ERROR] Errors: 
[ERROR]   TMailMailboxAppenderTest.appendShouldAddMessageToDesiredTeamMailbox:75 » Messaging
```

.block is likely the guilty one, yes. Please fix the unit test...

Arsnael (approved):

Arsnael:
Could that explain as well the degradation of perf we noticed on imap, as we switched to TMailLocalDelivery on preprod?

quantranhong1999:
> Could that explain as well the degradation of perf we noticed on imap, as we switched to TMailLocalDelivery on preprod?

I think no. We don't really use the mailet pipeline while doing IMAP perf test, right?

Arsnael:
Hmmm yes but still we send emails that coming back in the mailet pipeline and is maybe blocking cassandra which makes other operations towards Cassandra then longer? I don't know honestly

vttranlina (approved):
good job

chibenwa (approved):

quantranhong1999:
> TODO: run a performance test round to make sure com.datastax.oss.driver.api.core.DriverTimeoutException related to MailDispatcher is not existed.

good to me

PR #617, opened 2023-03-24
[OIDC] Demo with OIDC (lemon) set up and TeamMail flutter web frontend

chibenwa:


Arsnael:
Read it. No comment for now

PR #621, opened 2023-03-28
ISUE-588 Linagora service discovery

quantranhong1999:


tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/LinagoraServicesDiscoveryRoutesContract.scala
```diff
@@ -0,0 +1,108 @@
...
+      .extract()
+      .body()
+      .asString()
+
+    assertThatJson(response)
+      .isEqualTo(
+        s"""{
+           |	"linShareApiUrl": "https://linshare.linagora.com/linshare/webservice",
+           |	"linToApiUrl": "https://linto.ai/demo",
+           |	"linToApiKey": "apiKey",
```
chibenwa, line 48: This call should likely be authenticated, as details like linto API key may be sensible...

quantranhong1999: Yes, authentication has been done here. Test name `serviceDiscoveryShouldSucceedWhenUserIsAuthenticated` (basic auth in Before each BTW).

Maybe I would explicitly declare it for more readable.

chibenwa (approved):

Arsnael (approved):

PR #626, opened 2023-03-31
ISSUE-609 JMAP CalendarEvent/parse STATUS

quantranhong1999:
In a "VEVENT" calendar component, the "Organizer" can indicate that a meeting is tentative, confirmed, or cancelled.

RF:
https://www.rfc-editor.org/rfc/rfc8984.html#section-5.1.3 
https://www.rfc-editor.org/rfc/rfc5545#section-3.8.1.11


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/model/CalendarEventParse.scala
```diff
@@ -105,6 +106,24 @@ object CalendarFreeBusyStatusField extends Enumeration {
 
 case class CalendarFreeBusyStatusField(value: FreeBusyStatus) extends AnyVal
 
+object CalendarEventStatusField extends Enumeration {
+  type EventStatus = Value
+  val Confirmed = Value("confirmed")
+  val Cancelled = Value("cancelled")
+  val Tentative = Value("tentative")
```
chibenwa, line 113: Shouldn't this be implemented as an enum? (sealed trait + case objects ?)

quantranhong1999: Here I am using directly the enum syntax of Scala 2.xx: https://www.scala-lang.org/api/2.13.5/scala/Enumeration.html

Is it not less boilerplate than sealed trait + case objects?

chibenwa (approved):

Arsnael (approved):

PR #654, opened 2023-04-20
ISSUE-650 Some CalendarEvent/parse changes

quantranhong1999:


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/json/CalendarEventSerializer.scala
```diff
@@ -69,5 +69,20 @@ object CalendarEventSerializer {
...
+      .map(blobIdToCalendarEvent => blobIdToCalendarEvent._1 -> blobIdToCalendarEvent._2.transform {
+        case jsonObject: JsObject => JsSuccess(properties.filter(jsonObject))
+        case js => JsSuccess(js)
+      }.fold(_ => JsObject(Map(blobIdToCalendarEvent)), o => o)))
+
+    originalJson.transform((__ \ "parsed").json.prune).get
+      .transform((__).json.update(
+        __.read[JsObject].map(prunedParsed => prunedParsed ++ JsObject(Map("parsed" -> filteredParsed)))))
+      .get
+  }
```
quantranhong1999, line 87: I was struggling to filter the properties for calendar events here.
Tried something traditional to classic filtering in James but does not work (this case is a bit diff and a bit more complicated):
```
  def serializeCalendarEventResponse(calendarEventResponse: CalendarEventParseResponse, properties: Properties): JsValue =
    Json.toJson(calendarEventResponse)
      .transform((__ \ "parsed").json.update {
        case JsObject(underlying) => {
          val filteredObject: JsObject = JsObject(underlying
            .map(blobIdToCalendarEvent => (blobIdToCalendarEvent._1 -> blobIdToCalendarEvent._2.transform {
              case jsonObject: JsObject => JsSuccess(properties.filter(jsonObject))
              case js => JsSuccess(js)
            }.fold(_ => JsObject(underlying), o => o))))
          val tempJssuccess = JsSuccess(filteredObject)
          tempJssuccess
        }
        case jsValue => JsSuccess(jsValue)
      }).get
```
The `filteredObject` filtered the properties as expected, but the final result is not filtered...

The current solution works for properties filtering but it may need more clean/error handling (some tests failing and I did not have time to fix all of them yet).

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/LinagoraCalendarEventParseMethodContract.scala
```diff
@@ -1068,6 +1069,300 @@ trait LinagoraCalendarEventParseMethodContract {
...
+    assertThatJson(response)
+      .withOptions(new Options(IGNORING_ARRAY_ORDER))
+      .inPath("methodResponses[0]")
+      .isEqualTo(
+        s"""[
+           |    "CalendarEvent/parse",
+           |    {
+           |        "accountId": "$ACCOUNT_ID",
+           |        "parsed": {
+           |            "$blobId": {
```
chibenwa, line 1154:
```suggestion
           |        "parsed": {
           |            "$blobId": [{
```

Should be an array now no?

quantranhong1999: Yes, TODO in other commit

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/json/CalendarEventSerializer.scala
Arsnael: I'm not sure to get why so complicated to filter properties here... You seem to extract the parsed field then filter it separately and then retake the original json, prune the parsed field and replace with the filtered one... correct?

I don't really have time to put my hands into the code right now but just filtering directly fields in parsed is not possible? How is it because the properties fields are nested in the response?

quantranhong1999: > but just filtering directly fields in parsed is not possible?

Yes, I tried this at first with the code in the comment above. Even with some look with @vttranlina.

chibenwa (approved):

Arsnael (approved):

Arsnael:
Can you squash fixup and remove the wip from the second commit @quantranhong1999 ? :)

PR #661, opened 2023-04-25
CORS for APISIX

quantranhong1999:
Outside of the CORS setup scope, but I am trying to set up a smooth SSO flow for TMail FE + APISIX.

Currently, I can log in to TMail, however, can not do anything more with it because the wrong endpoint was called by TMail FE
image

TODO: check configuration regarding oidc endpoint, jmap prefix..

### DIFF
--- a/demo/apisix/conf/apisix.yaml
+++ b/demo/apisix/conf/apisix.yaml
@@ -12,6 +12,7 @@ routes:
     service_id: jmap_service_oidc
     methods:
       - POST
+      - OPTIONS
     plugins:
       proxy-rewrite:
         uri: /jmap
@@ -21,6 +22,7 @@ routes:
     service_id: jmap_service_oidc
     methods:
       - GET
+      - OPTIONS
     plugins:
       proxy-rewrite:
         uri: /jmap/session
@@ -30,6 +32,7 @@ routes:
     service_id: jmap_service_oidc
     methods:
       - GET
+      - OPTIONS
     plugins:
       proxy-rewrite:
         regex_uri:
@@ -41,6 +44,7 @@ routes:
     service_id: jmap_service_oidc
     methods:
       - POST
+      - OPTIONS
     plugins:
       proxy-rewrite:
         regex_uri:
@@ -52,6 +56,7 @@ routes:
     service_id: jmap_no_auth_service
     methods:
       - GET
+      - OPTIONS
     plugins:
       proxy-rewrite:
         uri: /.well-known/webfinger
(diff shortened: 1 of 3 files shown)


Arsnael (commented):
Or for .well-known-jmap apisix endpoint declaration, you could redirect maybe directly to jmap/session of tmail instead of .well-known/jmap endpoint? It's the same endpoint and user doesnt see what goes beyond apisix. Maybe can solve your issue?

quantranhong1999:
The SSO flow with basic usage is fine now:

[demo SSO TMail web.webm](https://user-images.githubusercontent.com/55171818/234464071-3dabf58a-5478-41f6-985e-09cc04c6ffc9.webm)

Arsnael (approved):
Good job :)

Arsnael:
Question: I don't see any OPTIONS call actually made from tmail-web in the network dev tools... is that normal? and are we sure it works then?

demo/apisix/conf/apisix.yaml
```diff
@@ -91,13 +91,14 @@ services:
...
-        client_id: "james_apisix"
-        client_secret: "james_apisix"
+        client_id: "james"
+        client_secret: "james"
         discovery: "http://sso.example.com/.well-known/openid-configuration"
         scope: "openid profile"
         bearer_only: true
         introspection_endpoint_auth_method: "client_secret_post"
         redirect_uri: "http://test.sso.example.com:8080/login-callback.html"
+        use_jwks: true
```
vttranlina, line 103: Q:
>  "TMail web will can not SSO with that client." 

how? Is it not redirect 302 when using the web browser?
if yes, so I am just curious why it works with Krakend, (I remember the Krakend does not redirect 302...)

quantranhong1999:
> Question: I don't see any OPTIONS call actually made from tmail-web in the network dev tools

Actually, there are (maybe not shown in the last demo).
[options request demonstration.webm](https://user-images.githubusercontent.com/55171818/234520017-f38f5835-b15a-4060-b5a4-bc13e469cd3a.webm)

demo/apisix/conf/apisix.yaml
vttranlina, line 103: If we choice jwks, and introspection is redundant ( the fact is it does not work with public =1), We should remove introspection relate

quantranhong1999: This would happen if we use introspection way (oidcRPMetaDataOptionsPublic = 0, therefore I switch to jwks mode.

Screenshot from 2023-04-25 16-54-43

> so I am just curious why it works with Krakend

Actually the Krakend in our demo use jwks mode too.

Arsnael (approved):

quantranhong1999:
Squashed.

vttranlina (approved):

PR #696, opened 2023-05-29
ISSUE-666 Contacts deletion: should remove the contact in domain contact

quantranhong1999:
IMO there should be no reason that other users still be suggested a deleted user contact. Some deployments may not depend on LSC/LDAP to sync the deletion for domain contact.


chibenwa (approved):
Good catch

Arsnael (approved):

PR #712, opened 2023-06-02
ISSUE-702 JMAP specifications for labels

quantranhong1999:


chibenwa (approved):

vttranlina (approved):

docs/modules/ROOT/pages/tmail-backend/jmap-extensions/jmapLabels.adoc
```diff
@@ -0,0 +1,160 @@
...
+The first two letters/numbers refer to red, the next two refer to green, and the last two refer to blue.
+The color values are defined in values between 00 and FF (0 to 255 in RGB).
+
+== Label/get
+
+Standard */get* method as described in Section 5.1, RFC8620.
+
+== Label/set
+
+Standard */set* method as described in Section 5.3, RFC8620.
```
Arsnael, line 34: Not completely. Only create and update are implemented, not destroy part for example :)

quantranhong1999: We may want/can do the destroyed part, no?

BTW I think we should not limit the feature at the spec level just because current implementation detail. e.g currently we find it hard to support destruction, but in the future, we can add that later with no limit from the spec?

chibenwa: As said with an extra level of integration we can even provide destroy ;-)

Arsnael: Alright :)

Arsnael (approved):

PR #768, opened 2023-07-28
ISSUE-765 Enable Quota RFC-9245 draft compatibility mode by default

### LINKED ISSUE #765: Enable Quota RFC-9245 compatibility mode.
Add by default the corresponding system property

quantranhong1999:
resolve #765


chibenwa (commented):
I was more thinking: put this by default in jvm.properties: makes it easy to drop ot once not needed...

quantranhong1999:
Would go for more flexible option.

PR #809, opened 2023-09-06
ISSUE-787 Push for Settings

### LINKED ISSUE #787: Push for Settings
**Original ticket:** https://github.com/linagora/tmail-backend/issues/630

**WHY**

Upon setting update, we need to push the change event to FCM

**HOW**

create SettingTypeName object (extend TypeName), push the event StateChangeEvent to the event bus after updating setting

**DoD**: pass contract test

Notes: need Settings/set method available.

quantranhong1999:
resolve #787 
TODO check someplace are seem to always push StateChangeEvent even no successful change e.g `FilterSetMethod`

### DIFF
--- a/tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/JmapSettingsSetMethod.scala
+++ b/tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/JmapSettingsSetMethod.scala
@@ -4,10 +4,15 @@ import com.linagora.tmail.james.jmap.json.JmapSettingsSerializer
 import com.linagora.tmail.james.jmap.method.CapabilityIdentifier.LINAGORA_SETTINGS
 import com.linagora.tmail.james.jmap.model.SettingsSet.OBJECT_ID
 import com.linagora.tmail.james.jmap.model.{SettingsSetError, SettingsSetRequest, SettingsSetResponse, SettingsSetUpdateRequest, SettingsUpdateResponse}
-import com.linagora.tmail.james.jmap.settings.JmapSettingsRepository
+import com.linagora.tmail.james.jmap.settings.{JmapSettingsRepository, SettingsTypeName}
 import eu.timepit.refined.auto._
-import javax.inject.Inject
+import javax.inject.{Inject, Named}
 import org.apache.james.core.Username
+import org.apache.james.events.Event.EventId
+import org.apache.james.events.EventBus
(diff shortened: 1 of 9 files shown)


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/JmapSettingsSetMethod.scala
```diff
@@ -97,7 +104,25 @@ class SettingsSetRequestSetMethod @Inject()(val jmapSettingsRepository: JmapSett
       oldState <- evaluateOldState(updateResults, currentStateIfAllUpdatesFailed)
       newState <- evaluateNewState(updateResults, currentStateIfAllUpdatesFailed)
       response = createResponse(invocation.invocation, request, updateResults, oldState, newState)
-    } yield InvocationWithContext(response, invocation.processingContext)
+    } yield {
+      dispatchSettingsChangeEvent(mailboxSession.getUser, oldState, newState).subscribe()
+
+      InvocationWithContext(response, invocation.processingContext)
```
chibenwa, line 110: Why not just flatMap?

I do not see why we need to be async here.

quantranhong1999: I thought about 2 points:
- sync makes the response slower (even dispatch event should be quick, but still slower)
- In the case RabbitMQ is down, then failure to dispatch an event to RabbitMQ forces the Settings/set to fail. Seems not that critical to me.

chibenwa: > sync makes the response slower (even dispatch event should be quick, but still slower)

IMO not a concern we should tackle here but more at the event bus layer.

Async here means anyway unbounded memory queue. Not sure that's a great idea...

> In the case RabbitMQ is down, then failure to dispatch an event to RabbitMQ forces the Settings/set to fail. Seems not that critical to me.

Again an event bus concern.

Event bus is supposed to save failure to dispatch into deadletter so I think that this concern is even already handled.

vttranlina (commented):
read it

hungphan227 (approved):

Arsnael (commented):
Nothing to add

Arsnael (approved):

Arsnael:
Needs rebase though

PR #815, opened 2023-09-15
Set up TTL on the `label_change` table

quantranhong1999:


tmail-backend/jmap/extensions-cassandra/src/main/scala/com/linagora/tmail/james/jmap/label/CassandraLabelChangesConfiguration.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.google.common.base.Preconditions
+import org.apache.commons.configuration2.Configuration
+import org.apache.james.util.DurationParser
+
+object CassandraLabelChangesConfiguration {
+  private val DEFAULT_TTL: Duration = Duration.ofDays(60)
+  val DEFAULT: CassandraLabelChangesConfiguration = CassandraLabelChangesConfiguration(DEFAULT_TTL)
+
+  def from(configuration: Configuration): CassandraLabelChangesConfiguration = {
+    val labelChangeTtl: Duration = Option(configuration.getString("label.change.ttl", null))
```
chibenwa, line 15: Where's the doc for this conf?

quantranhong1999: It seems TMail does not have a doc entry for additional configuration yet. I would add a docs for those.

upgrade-instructions.md
```diff
@@ -12,6 +12,27 @@ software documentation. Do not follow this guide blindly!
...
+
+From now on, we set a default 60 days (configurable) Cassandra time-to-live on the `label_change` table
+to reduce considered outdated changes' storage usage.
+
+If one want to keep the old behavior which is not using Cassandra time-to-live on those JMAP change tables, please set the
+following configurations in `cassandra.properties`:
+
+```
+label.change.ttl=0
+```
```
Arsnael, line 32: You don't precise that the value is in seconds. And should it be seconds or days btw?

quantranhong1999: It defaults to `second`. That is good enough for OPS IMO.

Arsnael: Maybe with the configuration doc? Or some examples? It's not trivial that if you put just a number it's in seconds and that you can add the unit of duration as well alongside it for example :)

Arsnael (approved):

chibenwa (approved):

PR #838, opened 2023-10-03
ISSUE-823 Webadmin task for inbox archival

quantranhong1999:


tmail-backend/webadmin/webadmin-mailbox/src/main/java/com/linagora/tmail/webadmin/archival/InboxArchivalService.java
```diff
@@ -0,0 +1,146 @@
...
+    private Predicate<MailboxMessage> messagesOlderThanArchiveDate(Date archiveDate) {
+        return mailboxMessage -> mailboxMessage.getSaveDate()
+            .orElse(mailboxMessage.getInternalDate())
+            .before(archiveDate);
+    }
+
+    private Mono<Void> archiveMessage(MailboxMessage mailboxMessage, MailboxSession mailboxSession, JmapSettings jmapSettings) {
+        return switch (jmapSettings.inboxArchivalFormat().toString()) {
+            case "yearly" -> yearlyArchive(mailboxMessage, mailboxSession);
+            case "monthly" -> monthlyArchive(mailboxMessage, mailboxSession);
```
chibenwa, line 107: Use constants

tmail-backend/webadmin/webadmin-mailbox/src/main/java/com/linagora/tmail/webadmin/archival/InboxArchivalService.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private Mono<Void> monthlyArchive(MailboxMessage mailboxMessage, MailboxSession mailboxSession) {
+        LocalDate savedDate = mailboxMessage.getSaveDate()
+            .orElse(mailboxMessage.getInternalDate())
+            .toInstant()
+            .atZone(ZoneId.of("UTC"))
+            .toLocalDate();
+        MailboxPath monthlyArchiveSubMailbox = MailboxPath.forUser(mailboxSession.getUser(),
+            String.format("Archive.%d.%s", savedDate.getYear(), savedDate.getMonth().getDisplayName(TextStyle.FULL, Locale.US)));
```
chibenwa, line 135: You can't infer the local for month display: `01` is waaaay better than `January` for my Mum for instance.

Let's do this based on the user local setting then? Or maybe front would be able to adjust the display?

tmail-backend/webadmin/webadmin-mailbox/src/main/java/com/linagora/tmail/webadmin/archival/InboxArchivalService.java
```diff
@@ -0,0 +1,146 @@
...
+
+        return moveMessage(mailboxMessage, mailboxSession, monthlyArchiveSubMailbox);
```
chibenwa, line 137: Moving messages one by one is likely sub-optimal. I suggest to group by target folder and try to limit the number of moves.

chibenwa (commented):
For this task, as well as the other cleanup tasks, could we plan on measuring their throughtput (message per second) on preprod?

+ add it in the task doc...

quantranhong1999:
> For this task, as well as the other cleanup tasks, could we plan on measuring their throughtput (message per second) on preprod?

Do you mean to measure smth like (cleaned messages number) / (task run time)?

chibenwa:
Yes.

I want to ensure scheduling this task won't blow up prods. Let's understand its performance characteristics and run time...

Having an idea in MSG / sec is IMO a good idea. A limited run on preprod should already give us an idea...

tmail-backend/webadmin/webadmin-mailbox/src/main/java/com/linagora/tmail/webadmin/archival/InboxArchivalService.java
quantranhong1999: > Let's do this based on the user local setting then?

I can think about some month display formats that can be done on the backend side:
  - number: 01 02 ... 12
  - full month name: January, February,...
  - Short month name: Jan, Feb,...
 
Less work on the frontend side.

In the case user changes the month display format, we can either:
- Adjust the display on the frontend side
- Or on backend side we listen to the Setting change event and rename the archived mailboxes.

chibenwa: The problem is that you cannot infer that the user speaks english!

chibenwa: Both solution you propose are OK and would be an enhancement.

Arsnael: I think numbers are enough IMO. It talks for everyone despite their language, let's keep it simple? :)

quantranhong1999: Yes. Let's keep it number then, and let frontend side manipulate the display if needed.

tmail-backend/webadmin/webadmin-mailbox/src/main/java/com/linagora/tmail/webadmin/archival/InboxArchivalService.java
quantranhong1999: not solved yet

quantranhong1999:
TODO adds successful users, and failed users to AdditionalInformation from https://github.com/linagora/tmail-backend/pull/839#issuecomment-1746586006, which seems useful too,

hungphan227 (approved):

Arsnael:
Read it nothing to add, looks good so far :)

tmail-backend/webadmin/webadmin-mailbox/src/main/java/com/linagora/tmail/webadmin/archival/InboxArchivalService.java
quantranhong1999: Seems not to be able to solve with the current API to me.

quantranhong1999:
Rebase on master with the latest fixup https://github.com/linagora/tmail-backend/pull/838/commits/c592c975720322c6394003d49838324f4d941058

vttranlina (approved):

Arsnael (approved):

PR #878, opened 2023-12-13
ISSUE-870 Enable mutable-content for APNS notification

quantranhong1999:
Was needed by mobile team to handle Notification Service Extension (an extension to modify notification of iOS)

cc @dab246


dab246:
IMO, We should also set it to the condition below
Screenshot 2023-12-13 at 11 22 29

quantranhong1999:
> IMO, We should also set it to the condition below

Yes, you are right. It was a quick fix and I missed that. Thanks for reviewing :-)

dab246 (approved):

hoangdat (approved):

Arsnael (approved):

PR #879, opened 2023-12-14
Update base docker image to the latest stable JRE

quantranhong1999:


Arsnael (approved):

pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <properties>
         <target.jdk>17</target.jdk>
-        <jib.base.image>eclipse-temurin:20-jre-jammy</jib.base.image>
+        <jib.base.image>eclipse-temurin:21-jre-jammy</jib.base.image>
```
vttranlina, line 15: `jib.base.image` is only used in `ci` maven profile
We also need to change the base image on default https://github.com/linagora/tmail-backend/blob/dd61469faac399640cc89204aa683c167b27ca08/tmail-backend/apps/distributed/pom.xml#L415

both distributed and memory
IMO, we can create property:
<jib.base.image.default> for that 

// TLDR
why we don't use a single `jib.base.image` for all,
because we will build the multi-platform image when ci profile. 
and we "fixed" `amd64` digest when default (it will help avoid re-fetch image for each jib building)

quantranhong1999: > and we "fixed" amd64 digest when default (it will help avoid re-fetch image for each jib building)

IMO we should have not hard code the digest of the image as Docker already handles the case. 

If I am not wrong, when we do `docker pull imageX`: docker will compare the local digest to the remote digest on Docker Hub. If the digest is the same -> no actual download image, otherwise docker would download the image.

quantranhong1999: The same behavior should be the same for JIB. As its doc mentions:

`
Note that, even after Jib fully cached a base image, Jib still connects to Docker Hub to check the image every time it runs (unless you pin to a specific base image using a SHA digest). This is to check if the cached base image is up-to-date.
`
rf: https://github.com/GoogleContainerTools/jib/blob/master/docs/default_base_image.md#docker-hub-download-rate-limit

So I expect:
- using image + tag -> Jib would only check the digest and not download the image if the digest is not changed.
- using image + tag + fixed digest -> Jib would not even go online to check and just use the local image.

IMO we should just use image + tag. Reasons:
- More consistent on base image updates for dev. Do not need to care about digest here and there. JIB just goes online and peek if any update on the image is already good IMO.
- The digest of the image can be changed remotely (e.g. override the old image with vunerability update). For example, I see [`eclipse-temurin:17-jre-jammy`](https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=17-jre-jammy) was updated 12 days ago even jre 17 has been released for a long time. Relying on a fixed digest could potentially leads to not found image issue + security problem IMO.

vttranlina: > If I am not wrong, when we do docker pull imageX: docker will compare the local digest to the remote digest on Docker Hub. If the digest is the same -> no actual download image, otherwise docker would download the image.

you are right, I just tested it local. 

I'm ok when remove digest

Arsnael (approved):

chibenwa (approved):
Perf ?

chibenwa:
Shouldn't we switch the target to 21 too while we are at it?

BTW we benefit from JRE improvments.

Arsnael:
> Shouldn't we switch the target to 21 too while we are at it?

We could, but in an other PR I think

vttranlina (approved):

chibenwa (approved):

pom.xml
vttranlina: Another point after unset digest , 
` 429 Too Many Requests` from dockerhub :( 

image

PR #891, opened 2024-01-10
ISSUE-887: No strict check on links in discovery ecosystem endpoint

Arsnael:


tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/LinagoraServicesDiscoveryRoutesContract.scala
```diff
@@ -42,7 +42,7 @@ trait LinagoraServicesDiscoveryRoutesContract {
     assertThatJson(response)
       .isEqualTo(
         s"""{
-           |	"linShareApiUrl": "https://linshare.linagora.com/linshare/webservice",
+           |	"linshareApiUrl": "https://linshare.linagora.com/linshare/webservice",
```
quantranhong1999, line 45: Maybe we need to notify the mobile team about the change here.

quantranhong1999 (approved):

chibenwa (approved):

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/LinagoraServicesDiscoveryRoutesContract.scala
Arsnael: No need really... you can put whatever as a key now.

Or maybe you prefer to make that change on the configuration sample file to avoid confusion?

quantranhong1999 (approved):

Issue #910, opened 2024-01-30
[Postgresql] Implement PostgresFirebaseSubscriptionRepository


hungphan227:
## How
- Table structure:
```sql
CREATE TABLE firebase_subscription (
  user TEXT,
  device_client_id TEXT,
  id UUID,
  expires TIMESTAMP,
  fcm_token TEXT,
  types TEXT[],
  PRIMARY KEY (user, device_client_id)
);
```
- need RLS

## DoD
Pass contract

Arsnael:
DAO + PostgresFirebaseSubscriptionRepository + Guice bindings

quantranhong1999:
https://github.com/linagora/tmail-backend/pull/956

PR #956, opened 2024-03-07
ISSUE-910 Implement PostgresFirebaseSubscriptionRepository

quantranhong1999:
resolve https://github.com/linagora/tmail-backend/issues/910

### DIFF
--- a/tmail-backend/jmap/extensions-postgres/src/main/java/com/linagora/tmail/james/jmap/firebase/PostgresFirebaseSubscriptionDAO.java
+++ b/tmail-backend/jmap/extensions-postgres/src/main/java/com/linagora/tmail/james/jmap/firebase/PostgresFirebaseSubscriptionDAO.java
@@ -1,0 +1,145 @@
+package com.linagora.tmail.james.jmap.firebase;
+
+import static com.linagora.tmail.james.jmap.firebase.PostgresFirebaseModule.FirebaseSubscriptionTable.DEVICE_CLIENT_ID;
+import static com.linagora.tmail.james.jmap.firebase.PostgresFirebaseModule.FirebaseSubscriptionTable.EXPIRES;
+import static com.linagora.tmail.james.jmap.firebase.PostgresFirebaseModule.FirebaseSubscriptionTable.FCM_TOKEN;
+import static com.linagora.tmail.james.jmap.firebase.PostgresFirebaseModule.FirebaseSubscriptionTable.FCM_TOKEN_UNIQUE_CONSTRAINT;
+import static com.linagora.tmail.james.jmap.firebase.PostgresFirebaseModule.FirebaseSubscriptionTable.ID;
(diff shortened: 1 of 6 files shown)


Arsnael (approved):

tmail-backend/jmap/extensions-postgres/src/main/java/com/linagora/tmail/james/jmap/firebase/PostgresFirebaseSubscriptionDAO.java
```diff
@@ -0,0 +1,145 @@
...
+            CollectionConverters.asScala(extractTypes(record)).toSeq());
+    }
+
+    private FirebaseSubscriptionExpiredTime toExpires(Record record) {
+        return new FirebaseSubscriptionExpiredTime(OFFSET_DATE_TIME_ZONED_DATE_TIME_FUNCTION.apply(record.get(EXPIRES)));
+    }
+
+    private Set<TypeName> extractTypes(Record record) {
+        return Arrays.stream(record.get(TYPES))
+            .map(string -> typeStateFactory.parse(string).right().get())
```
chibenwa, line 142: Do we handle "no longer existing types" stored on the DB? 
How?
What's the impact?
What's the alternative?

quantranhong1999: > Do we handle "no longer existing types" stored on the DB?
How?

It is likely not the case as of today. (many place in our code as well)

> What's the impact?

Likely:
- FirebaseSubscription/get and PushSubscription/get would fail
- Pushing user notifications via Push gateway / FCM would fail because we can not deserialize the "no longer existing types".

> What's the alternative?

Be more lenient parsing the types I guess.

But is that a soon viable issue when we give up a supported type?

chibenwa: Let's not fix that as off now but let's document the case in a ticket.

IMO being lenient + WARN log would be the bet when reading our data store.

Strict enforcement however on what is set by our clients.

chibenwa: Can you open the ticket?

quantranhong1999: > Can you open the ticket?

Yes. let me.

quantranhong1999: https://github.com/linagora/tmail-backend/issues/958

hungphan227 (approved):

PR #968, opened 2024-03-12
[FIX] Remove module chooser for deleted message vault for TMail Postgres app

quantranhong1999:
cf TMail Distributed app: deleted message vault was always loaded e.g. for JMAP EmailRecoveryAction method


quantranhong1999:
> Or maybe we want the possibility to choose but we just messed it up in the code?

Nope. This is to align to the same behavior as our TMail Distributed app: always load deleted messages vault module (likely we want to always have JMAP EmailRecoveryAction method).

Unless we should have not wanted it in the TMail Distributed app at first.

vttranlina (approved):
good catch

chibenwa (approved):

PR #1022, opened 2024-04-23
ISSUE-941 Clean redis event bus upon james crash

quantranhong1999:
TODO:
- [ ] Cleanup logic + Unit tests for the task
- [ ] Guice binding + Integration tests
- [ ] Documentation

### DIFF
--- a/tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/CleanRedisEventBusService.java
+++ b/tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/CleanRedisEventBusService.java
@@ -1,0 +1,147 @@
+package org.apache.james.events;
+
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
+
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+
+import org.apache.james.util.MDCStructuredLogger;
+import org.apache.james.util.ReactorUtils;
+import org.apache.james.util.StructuredLogger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import io.lettuce.core.KeyScanArgs;
+import io.lettuce.core.KeyScanCursor;
+import io.lettuce.core.ScanCursor;
+import io.lettuce.core.api.reactive.RedisKeyReactiveCommands;
+import io.lettuce.core.api.reactive.RedisSetReactiveCommands;
+import io.lettuce.core.pubsub.api.reactive.RedisPubSubReactiveCommands;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public class CleanRedisEventBusService {
+    public static class Context {
+        private final AtomicLong totalBindings;
+        private final AtomicLong danglingBindings;
+        private final AtomicLong cleanedBindings;
+
+        public Context() {
+            this.totalBindings = new AtomicLong();
+            this.cleanedBindings = new AtomicLong();
+            this.danglingBindings = new AtomicLong();
+        }
+
+        void increaseScannedBindingsCount() {
+            totalBindings.incrementAndGet();
+        }
+
+        void increaseDanglingBindingsCount(long count) {
+            danglingBindings.addAndGet(count);
+        }
+
+        void increaseCleanedBindingsCount(long count) {
+            cleanedBindings.addAndGet(count);
+        }
+
+        public long getTotalBindings() {
+            return totalBindings.get();
+        }
+
+        public long getDanglingBindings() {
+            return danglingBindings.get();
+        }
+
+        public long getCleanedBindings() {
+            return cleanedBindings.get();
+        }
+    }
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CleanRedisEventBusService.class);
+
+    private final RedisSetReactiveCommands<String, String> redisSetCommands;
+    private final RedisPubSubReactiveCommands<String, String> redisPubSubCommands;
+    private final RedisKeyReactiveCommands<String, String> redisKeyCommands;
+    private final RoutingKeyConverter routingKeyConverter;
+    private Context context;
+
+    @Inject
+    @Singleton
+    public CleanRedisEventBusService(RedisEventBusClientFactory redisEventBusClientFactory,
+                                     RoutingKeyConverter routingKeyConverter) {
+        this.redisSetCommands = redisEventBusClientFactory.createRedisSetCommand();
+        this.redisPubSubCommands = redisEventBusClientFactory.createRedisPubSubCommand();
+        this.redisKeyCommands = redisEventBusClientFactory.createRedisKeyCommand();
+        this.routingKeyConverter = routingKeyConverter;
+    }
+
+    public Mono<Void> cleanUp() {
+        this.context = new Context();
+
+        return listActiveRedisChannels()
+            .collectList()
+            .flatMap(activeChannels -> listRedisSetKeys()
+                .filter(this::isEventBusBindingKey)
+                .flatMap(key -> listElementsOfSet(key)
+                        .doOnNext(channel -> context.increaseScannedBindingsCount())
+                        .filter(channel -> !activeChannels.contains(channel))
+                        .collectList()
+                        .doOnNext(inactiveChannels -> context.increaseDanglingBindingsCount(inactiveChannels.size()))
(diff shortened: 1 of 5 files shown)


tmail-backend/webadmin/webadmin-redis-event-bus/src/main/java/com/linagora/tmail/webadmin/redis/CleanRedisEventBusRoutes.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.james.webadmin.Routes;
+import org.apache.james.webadmin.tasks.TaskFromRequest;
+import org.apache.james.webadmin.utils.JsonTransformer;
+
+import com.google.common.base.Preconditions;
+
+import spark.Request;
+import spark.Service;
+
+public class CleanRedisEventBusRoutes implements Routes {
```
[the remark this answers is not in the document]
quantranhong1999: Do you mean something like this?
image

The idea sounds good to me.

tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/CleanRedisEventBusService.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private Flux<String> listActiveRedisChannels() {
+        return redisPubSubCommands.pubsubChannels();
+    }
+
+    private Flux<String> listRedisSetKeys() {
+        return redisKeyCommands.scan(ScanCursor.INITIAL, KeyScanArgs.Builder.type("set"))
+            .flatMapIterable(KeyScanCursor::getKeys);
+    }
```
quantranhong1999, line 110: Testing on staging, I spotted there is a bug where Redis only returns 10/11 elements as a first batch. I would need to do some iteration to get all the elements here (with a test)

tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/CleanRedisEventBusService.java
```diff
@@ -0,0 +1,147 @@
...
+import io.lettuce.core.KeyScanCursor;
+import io.lettuce.core.ScanCursor;
+import io.lettuce.core.api.reactive.RedisKeyReactiveCommands;
+import io.lettuce.core.api.reactive.RedisSetReactiveCommands;
+import io.lettuce.core.pubsub.api.reactive.RedisPubSubReactiveCommands;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public class CleanRedisEventBusService {
+    public static class Context {
```
chibenwa, line 28: This "context" is a remain from the previous webadmin version of the code, and unused here.

Remove it?

tmail-backend/guice/distributed/src/main/java/com/linagora/tmail/event/RabbitMQAndRedisEventBusModule.java
```diff
@@ -81,10 +83,15 @@ sender, receiverProvider, eventSerializer, retryBackoffConfiguration, new Routin
...
     @ProvidesIntoSet
-    InitializationOperation workQueue(RabbitMQAndRedisEventBus instance) {
+    InitializationOperation workQueue(RabbitMQAndRedisEventBus instance, CleanRedisEventBusService cleanRedisEventBusService) {
         return InitilizationOperationBuilder
             .forClass(RabbitMQAndRedisEventBus.class)
-            .init(instance::start);
+            .init(() -> {
+                instance.start();
+                cleanRedisEventBusService.cleanUp()
+                    .subscribeOn(ReactorUtils.BLOCKING_CALL_WRAPPER)
```
chibenwa, line 92: Why blocvking call wrapper?

tmail-backend/guice/distributed/src/main/java/com/linagora/tmail/event/RabbitMQAndRedisEventBusModule.java
```diff
@@ -81,10 +83,15 @@ sender, receiverProvider, eventSerializer, retryBackoffConfiguration, new Routin
...
+                    .subscribe();
```
chibenwa, line 93: Why?

 -> Is it ok to start the redis event bus while the previous one isn't well clean up?
 
 -> How long does this take? Is it necessary to do it asynchronously?
 
 Doing this with a `.block` in the init sequence looks safer and simpler to me.
 
 Thoughts?

quantranhong1999: > Is it ok to start the redis event bus while the previous one isn't well clean up?

Yes, it is. Publish messages to the dangling bindings would not fail. I tested in the previous commits.

> How long does this take?

I can not say how long yet. Maybe I can measure it on staging tomorrow. My guess: a few seconds if the number of bindings is big.

> Is it necessary to do it asynchronously?

My idea when I put it asynchronous is to avoid causing more startup time for TMail, potentially reducing the downtime too. Given that JVM startup time is long already and especially TMail/James is fatter over time.

I am not sure how long the cleanup takes, especially when the data scales like millions of mailboxes, so I thought it would be safer (for startup time) to do it async.

> Doing this with a .block in the init sequence looks safer and simpler to me.

ATM I do log the result of the cleanup, with the number of bindings cleaned... So I think it is "safe" enough to know how the cleanup ran.

tmail-backend/guice/distributed/src/main/java/com/linagora/tmail/event/RabbitMQAndRedisEventBusModule.java
quantranhong1999: To avoid any cleanup code could run on the TMail startup thread and make the startup time slower.

tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/CleanRedisEventBusService.java
```diff
@@ -0,0 +1,147 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(CleanRedisEventBusService.class);
+
+    private final RedisSetReactiveCommands<String, String> redisSetCommands;
+    private final RedisPubSubReactiveCommands<String, String> redisPubSubCommands;
+    private final RedisKeyReactiveCommands<String, String> redisKeyCommands;
+    private final RoutingKeyConverter routingKeyConverter;
+    private Context context;
+
+    @Inject
+    @Singleton
```
vttranlina, line 73: Maybe we don't need "bean" for it. 
We only use it one time when start up

tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/CleanRedisEventBusService.java
```diff
@@ -0,0 +1,147 @@
...
+                .filter(this::isEventBusBindingKey)
+                .flatMap(key -> listElementsOfSet(key)
+                        .doOnNext(channel -> context.increaseScannedBindingsCount())
+                        .filter(channel -> !activeChannels.contains(channel))
+                        .collectList()
+                        .doOnNext(inactiveChannels -> context.increaseDanglingBindingsCount(inactiveChannels.size()))
+                        .flatMap(deleteBindingsToInactiveChannels(key)),
+                    ReactorUtils.DEFAULT_CONCURRENCY)
+                .then())
+            .doOnSuccess(any -> mdcLogWithContext()
```
vttranlina, line 97: Should handle case errors because of concurrency. 
eg: I scale up to 3 instances james at a time. 
Then maybe instance B can throws an error when trying delete th key that was done by instance A

tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/CleanRedisEventBusService.java
quantranhong1999: Ok. It would be better for memory as the bean is likely a long-lived object.

quantranhong1999:
I tested this on staging and it seems useful.

- A first cleanup: `{"danglingBindings":"4975","totalBindings":"4975","cleanedBindings":"4975"}`
  -> Cleaned something

- Cleanup during a performance test: `{"danglingBindings":"0","totalBindings":"4999","cleanedBindings":"0"}`
  -> Do not clean the being used bindings

- Cleanup after a performance test: `{"danglingBindings":"0","totalBindings":"0","cleanedBindings":"0"}`
  -> The bindings are well cleaned up by James. The cleanup service has nothing to do.

chibenwa (approved):

Arsnael (approved):

PR #1028, opened 2024-05-07
ISSUE-871 Backend should clean expired FCM subscriptions

quantranhong1999:


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/firebase/CleanExpiredFcmSubscriptionService.scala
```diff
@@ -0,0 +1,36 @@
...
+import reactor.core.scala.publisher.{SFlux, SMono}
+
+object CleanExpiredFcmSubscriptionService {
+  private val LOGGER: Logger = LoggerFactory.getLogger(classOf[CleanExpiredFcmSubscriptionService])
+}
+
+class CleanExpiredFcmSubscriptionService @Inject()(subscriptionRepository: FirebaseSubscriptionRepository,
+                                                   usersRepository: UsersRepository,
+                                                   clock: Clock) extends Startable {
+  def cleanUp(): SMono[Int] = {
```
chibenwa, line 22: If I do not reboot james, the cleanup is not done...

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/FirebaseSubscriptionGetMethod.scala
```diff
@@ -45,13 +49,21 @@ class FirebaseSubscriptionGetMethod @Inject()()(val serializer: FirebaseSubscrip
...
             .map(_.get)
             .toSet.asJava))
-    }
+    })
+      .flatMap(subscription => {
+        if (isInThePast(subscription.expires, clock)) {
+          SMono(repository.revoke(username, subscription.id))
+            .`then`(SMono.empty)
+        } else {
+          SMono.fromCallable(() => subscription)
```
chibenwa, line 66:
```suggestion
          SMono.just(subscription)
```

quantranhong1999: `SMono.just` is an eager evaluation. I want a lazy evaluation here.

chibenwa: Isnt subscription already evaluated?

quantranhong1999: I mean the `SMono` variable would be eagerly created even if the flow does not fall into the else block, which is useless, no?

chibenwa: I think there is not difference for serving a pojo between the two.

Arsnael (approved):

PR #1032, opened 2024-05-13
[ENHANCEMENT] Avoid copy of the mime message in calendar replies

chibenwa:
Cc @vttranlina


vttranlina (approved):
I tried to do the same in past, but I do not remember why  in the last test not passed

quantranhong1999 (approved):

Arsnael (approved):

PR #1034, opened 2024-05-14
[FIX] FCM: do not flag background update (`content-available=1`) for EmailDelivery notifications

quantranhong1999:
cf: https://github.com/linagora/tmail-backend/commit/43c49615a06edb850fdd7d4b35fbd05873dfc6af#r141977107

cc @tddang-linagora I deployed this fix on staging.

### DIFF
--- a/tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/firebase/FirebasePushClient.java
+++ b/tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/firebase/FirebasePushClient.java
@@ -115,7 +115,6 @@ public class FirebasePushClient {
                     .setTitle("Twake Mail")
                     .setBody("You have new message")
                     .build())
-                .setContentAvailable(true)
                 .setMutableContent(true)
                 .build();
         }


tddang-linagora:
@quantranhong1999 please refer to this PR's doc for future confirmation & issue
- https://github.com/linagora/tmail-flutter/pull/2860

tddang-linagora:
Suggest: add "for type EmailDelivery" in the title for clarification

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/firebase/FirebasePushClient.java
```diff
@@ -115,7 +115,6 @@ private Aps buildApsDictionary(FirebasePushRequest pushRequest) {
                     .setTitle("Twake Mail")
```
quantranhong1999, line 115:
```suggestion
                    .setTitle("Twake Mail for type EmailDelivery")
```

Do you mean adding something like this, and just to ease the testing on staging?

tddang-linagora: I meant title for this PR :)

tddang-linagora: as it only affects notification payloads that contain EmailDelivery

dab246: IMO, the default title should only be `Twake Mail` to avoid notification cases from being modified as desired by the client.

dab246 (approved):

chibenwa (approved):

PR #1035, opened 2024-05-14
PGSQL Update James SHA-1

Arsnael:


Arsnael:
https://james-jenkins.lin-saas.com/job/Tmail%20build/job/PR-1035/2/testReport/junit/com.linagora.tmail.deployment/PostgresImapAndSmtpTest/validateDeployment/

Investigating... Latest work is abot pool connection. Seems to alter the server responses order somehow on the deployment tests.

Arsnael:
> Investigating... Latest work is abot pool connection. Seems to alter the server responses order somehow on the deployment tests.

I dont get why on imap mpt tests there is some tests with flag updates on multiple messages and it's fine, while deployment tests somehow the response order is altered.

Does not happen before the migration to pg pool connection. @hungphan227 any ideas perhaps?

hungphan227:
I have reverted james sub module back to a commit before pool:

image

Then ran the test, it still fail:
```

org.apache.james.mpt.protocol.ProtocolSession$InvalidServerResponseException: 
Location: /org/apache/james/imap/scripts/ValidateDeployment.test:299
LastClientMsg: A15 STORE 1:2  +FLAGS (\FLAGGED)
Expected: '\* 1 FETCH \(FLAGS \(\\Flagged \\Recent\)\)'
Actual   : '* 2 FETCH (FLAGS (\Flagged \Recent))'
```

chibenwa:
@hungphan227 this looks interesting....

 - Does only the PGSQL backend fails or all backend fails?
 - I suggest you use `git bisect` to find out the one commit that introduced the bug...

Arsnael:
> I suggest you use git bisect to find out the one commit that introduced the bug...

Actually did... For the commit that commented Hung, it was working for me somehow. The work on the pool was my bisect bad. 

@hungphan227 the deployment tests are also based on the local image built with jib during compilation. Just to be sure, when you reverted, did you recompile the all project and rebuilt the image with jib before relaunching the test? If not you likely retested on this work.

chibenwa:
Do we have a way to reproduce the failure in James code base?

This would make it way easier to treat and manage the issue...

Arsnael:
> Do we have a way to reproduce the failure in James code base?

As said I tried using the similar scripts with mpt imap tests. In PostgresSelectedState, based on SelectedState test, some tests rely on the Store.test script, that has a store operation a bit similar. But it is green. 

Tried as well tinkering the script a bit storing flags in more messages (like storing the Answered flag from message 1:4) but still green...

Thought maybe then it's TMail? But The binding of the pool is done in PostgresCommonModule that is bound already in tmailPostgresServer as well...

Arsnael:
fix => https://github.com/apache/james-project/pull/2259

Was painful to do debugging by add logs everywhere in that part of the code, rebuilding the app and running the tests checking the logs... But I pinned it down to that line. concatMap instead of flatMap at that line ensure we keep the right uid order for the fetch responses of the store imap command

Arsnael:
Fix merged, let's see if the CI now is happy

Arsnael:
Can reproduce locally. Some things are wrong with postgres integration tests on tmail now

hungphan227:
I have fixed some tests in TeamMailboxesContract. PR: https://github.com/apache/james-project/pull/2267
However, there is still one test I could not fix: webSocketShouldPushTeamMailboxStateChanges

Arsnael:
> I have fixed some tests in TeamMailboxesContract. PR: https://github.com/apache/james-project/pull/2267

Are you sure this is not some isolation issue in the tests instead?

hungphan227:
> > I have fixed some tests in TeamMailboxesContract. PR: [apache/james-project#2267](https://github.com/apache/james-project/pull/2267)
> 
> Are you sure this is not some isolation issue in the tests instead?

"Isolation issue" -> what do you mean?

Arsnael:
> "Isolation issue" -> what do you mean?

That I think you get some key violation constraints because of entries already present and not cleaned up well from other tests.

hungphan227:
> > "Isolation issue" -> what do you mean?
> 
> That I think you get some key violation constraints because of entries already present and not cleaned up well from other tests.

That is not possible. We always drop all tables after each test

Arsnael:
> That is not possible. We always drop all tables after each test

Last commit, works for me

quantranhong1999:
> I have fixed some tests in TeamMailboxesContract. PR: https://github.com/apache/james-project/pull/2267

cc @hungphan227 can you check if it is the case for label change (TMail side)?

hungphan227:
> > I have fixed some tests in TeamMailboxesContract. PR: [apache/james-project#2267](https://github.com/apache/james-project/pull/2267)
> 
> cc @hungphan227 can you check if it is the case for label change (TMail side)?

If Rene commit works, my PR is not needed. However, I did not see any real difference made in Rene commit. Really surprised if it actually work

quantranhong1999:
> If Rene commit works, my PR is not needed.

Is upsert supported by the Cassandra implementation?

hungphan227:
> > If Rene commit works, my PR is not needed.
> 
> Is upsert supported by the Cassandra implementation?

Yes

quantranhong1999:
> Yes

Then we still need your upsert fix IMO.

Arsnael:
Ok with that

hungphan227:
> > Yes
> 
> Then we still need your upsert fix IMO.

Does Cassandra impl supporting upsert mean that postgres impl should support too?

Arsnael:
All implems should have similar behavior on this kind of operation

Arsnael:
Very unstable...

Sometimes some tests pass and sometimes not. My guess is there is a slight chance maybe with the email and mailbox change that there is two requests on different pool threads being executed in parallel with the same state, thus why sometimes postgres throws an exception on duplicated primary key entry?

A wild guess... Hard to debug honestly, as most of the time when I try to play a test alone it's green u_u

hungphan227:
> Very unstable...
> 
> Sometimes some tests pass and sometimes not. My guess is there is a slight chance maybe with the email and mailbox change that there is two requests on different pool threads being executed in parallel with the same state, thus why sometimes postgres throws an exception on duplicated primary key entry?
> 
> A wild guess... Hard to debug honestly, as most of the time when I try to play a test alone it's green u_u

Which test? Websocket?

Arsnael:
Not always the same honestly. Once I even got a green build, to tell you how unstable it is

quantranhong1999:
> Not always the same honestly. Once I even got a green build, to tell you how unstable it is

Hi. What is the name of the unstable test?

Arsnael:
You launch the postgres-jmap-integration-tests suite and you will see locally.

Tests failing recurrently for me are :
- `pushShouldSupportFilterTypeNameAndFilterStateWhenDataTypesAreNull` from LinagoraFilterStateChangeTest
- `webSocketShouldPushTeamMailboxStateChanges` from TeamMailboxesContract

But for example on my run now it seems it gets stuck on contactShouldBeIndexedWhenMailing in LinagoraContactAutocompleteMethodContract

I saw other tests sometimes red... and those tests sometimes being green.

So enjoy :)

Arsnael:
Problem is I had random errors on other tests, even locally. More than one test is unstable IMO

quantranhong1999:
> Problem is I had random errors on other tests, even locally. More than one test is unstable IMO

Hmm BTW unstable is the nature of those JMAP ITs IMO.

vttranlina:
I also get the unstable on `webSocketShouldPushTeamMailboxStateChanges` 
But the latest postgres is passed, it is unstable when commit upgrade james sha1

quantranhong1999:
I try to disable the `webSocketShouldPushTeamMailboxStateChanges` test here: https://github.com/linagora/tmail-backend/pull/1091

Let's see...

vttranlina:
The issue comes from `PoolBackedPostgresConnectionFactory` 
Commit: https://github.com/apache/james-project/commit/435600614fcfbc9bb2f4dccfea1b43a5069bd5a6#diff-29c9eefc2a62e1248232ac97ddadeaab789ffb85b10351e13e18aaa4e26c2195 (the change of `PostgresCommonModule.java`)

I tried to revert it to `SinglePostgresConnectionFactory` and `webSocketShouldPushTeamMailboxStateChanges` passed.

It looks like an issue related to what we did with Reactor (not logic issue),
the `StateChangeListener` already emits 2 messages, but the client subscriber receives only one.

Arsnael:
Thanks for the help team, that should do it :)

vttranlina:
Note that my fixup commit just makes test `webSocketShouldPushTeamMailboxStateChanges` not hanging
It still unstable 

We can @disable it as Quan's suggestion

Arsnael:
> It still unstable

Is it? Seemed okish to me :)

Arsnael:
When I said other tests were unstable and not just the websocket one... https://james-jenkins.lin-saas.com/job/Tmail%20build/job/PR-1035/10/

><

quantranhong1999 (approved):

hungphan227 (approved):

PR #1053, opened 2024-05-16
Fix case inserting existed blob id in PostgresBlobIdListDAO

hungphan227:


quantranhong1999 (approved):

chibenwa (commented):
Do we have a corresponding test case?

Arsnael (commented):
Read it, agree that if we could have a test for it it would be better

Arsnael (approved):

chibenwa (approved):

PR #1079, opened 2024-05-31
Update James SHA-1

Arsnael:
Trying to add quickly for the release the OOM IMAP COPY fix from @chibenwa


chibenwa:
Thanks for the initiative <3

PR #1104, opened 2024-06-20
POC Event bus notifications using Postgres Notify/Listen

quantranhong1999:
State: pass contract tests

TODO: Guice binding and aggregate a Docker image and performance test.

However, there is a scale issue with Postgres NOTIFY/LISTEN: 1 subscriber uses 1 dedicated connection -> we have a problem scaling the topic number e.g. a few hundred mailboxes would overload the default maximum 100 connections of Postgres (not saying about the pool max connections yet).

Would you like to continue with the performance test @chibenwa ?


chibenwa:
> Would you like to continue with the performance test @chibenwa ?

Does this translate to "one pg connection per event bus"?

If so it is not that bad IMO....

quantranhong1999:
> Does this translate to "one pg connection per event bus"?

Yes. Indeed it is not that bad.

chibenwa:
> Would you like to continue with the performance test @chibenwa ?

As explained with the ticket I wish we do not spend to much time on this POC but I am faily interested by the outcome so I am Ok with one man day testing this :-p

quantranhong1999:
Overall, the performance is good, not much different from the full RabbitMQ event bus one.

## IMAP performance
image

## JMAP performance
image


## Metrics
```
postgres_register{quantile="0.5",} 0.001835007
postgres_register{quantile="0.75",} 0.003358719
postgres_register{quantile="0.95",} 0.318767103
postgres_register{quantile="0.98",} 0.406847487
postgres_register{quantile="0.99",} 0.47395635100000005
postgres_register{quantile="0.999",} 0.7381975030000001
postgres_register_count 21083.0

postgres_dispatch{quantile="0.5",} 0.002932735
postgres_dispatch{quantile="0.75",} 0.004292607
postgres_dispatch{quantile="0.95",} 0.011993087000000001
postgres_dispatch{quantile="0.98",} 0.041156607000000005
postgres_dispatch{quantile="0.99",} 0.381681663
postgres_dispatch{quantile="0.999",} 0.809500671
postgres_dispatch_count 228676.0

postgres_unregister{quantile="0.5",} 0.0017121270000000001
postgres_unregister{quantile="0.75",} 0.0023756790000000003
postgres_unregister{quantile="0.95",} 0.007962623
postgres_unregister{quantile="0.98",} 0.022151167000000003
postgres_unregister{quantile="0.99",} 0.041156607000000005
postgres_unregister{quantile="0.999",} 0.33344716700000004
postgres_unregister_count 21083.0
```

The speed is not as fast as Redis, but OK IMO.

tmail-backend/event-bus-postgres/src/main/java/org/apache/james/events/TMailEventDispatcher.java
```diff
@@ -0,0 +1,223 @@
...
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+import reactor.rabbitmq.BindingSpecification;
+import reactor.rabbitmq.ExchangeSpecification;
+import reactor.rabbitmq.OutboundMessage;
+import reactor.rabbitmq.Sender;
+import reactor.util.function.Tuples;
+import reactor.util.retry.Retry;
+
+public class TMailEventDispatcher {
```
Arsnael, line 45: Totally understandable for the POC, but maybe we should try to refactor and generify something here with the existing TMailEventDispatcher with redis if we want to merge this work?

Not sure how difficult would that be though

tmail-backend/event-bus-postgres/src/main/java/org/apache/james/events/KeyChannelMessage.java
```diff
@@ -0,0 +1,33 @@
+package org.apache.james.events;
+
+import org.apache.commons.lang3.StringUtils;
+
+public record KeyChannelMessage(EventBusId eventBusId, String routingKey, String eventAsJson) {
```
Arsnael, line 5: I think that code seems very similar to existing KeyChannelMessage for redis pub sub and could be generified in case this work gets merged maybe too?

Arsnael (commented):
LGTM overall, just a few questions and remarks if we should refactor/polish some parts if we decide to merge this to the postgresql branch (@chibenwa seems to be in favor of that work?)

Also if we merge this, would be nice to have the documentation part with it too, stating well what @chibenwa said above, that we advise to use this just with very small deployments :)

tmail-backend/event-bus-postgres/src/main/java/org/apache/james/events/KeyChannelMessage.java
quantranhong1999: Yes, indeed this class is generic.

quantranhong1999:
> @chibenwa seems to be in favor of that work?

IMO unless we can replace all our queues (mail queues for example) to fully Postgres, then it is worth the merge. With only the event bus notifications using Postgres while mail queues still rely on RabbitMQ, I do not see how it would benefit small deployment (more complicated, fragment while not traded for better performance?).

chibenwa:
> IMO unless we can replace all our queues (mail queues for example) to fully Postgres, then it is worth the merge. With only the event bus notifications using Postgres while mail queues still rely on RabbitMQ, I do not see how it would benefit small deployment (more complicated, fragment while not traded for better performance?).

100% agree.

Arsnael:
Alright let's stop the experiment here then.

Thanks @quantranhong1999 for taking the patience and time to demonstrate this case :)

PR #1114, opened 2024-07-02
ISSUE-1090 TeamMailboxMember/set JMAP method

hungphan227:


tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/TeamMailboxMemberSetMethodContract.scala
```diff
@@ -0,0 +1,716 @@
...
+      .body("methodResponses[0]", jsonEquals(
+        s"""[
+           |    "TeamMailboxMember/set",
+           |    {
+           |        "accountId": "29883977c13473ae7cb7678ef767cbfbaffc8a44a6e463d971d23a65c1dc4af6",
+           |        "updated": {},
+           |        "notUpdated": {
+           |            "${teamMailbox.asString()}": {
+           |                "type": "invalidPatch",
+           |                "description": "Wrong teamMailboxName ${teamMailbox.asString()}"
```
chibenwa, line 503: We should return "not found" in order not to disclose the existance on the team mailbox.

I did not find the test were we try to add a member for a valid team mailbox name that do not exist.

Also the message is confusing here: after all the name is valid...

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+        invalidTeamMailboxNameFailureResult.concat(teamMailboxMemberSetResults.filter(setResult => setResult.notUpdated.nonEmpty))
+          .map(setResult => setResult.notUpdated.get.teamMailboxName -> setResult.notUpdated.get.error).toMap))
+  }
+
+  private def updateTeamMailboxMembers(username: Username,
+                                   teamMailbox: TeamMailbox,
+                                   mapMemberNameToRole: Map[TeamMailboxMemberName, Option[TeamMailboxMemberRoleDTO]]): SMono[TeamMailboxMemberSetResult] = {
+    SFlux(teamMailboxRepository.listMembers(teamMailbox))
+      .collectMap(member => member.username, member => member)
+      .flatMap(mapExistedMembers => mapExistedMembers.get(username) match {
```
chibenwa, line 72: s/mapExistedMembers/members/

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+          case Some(teamMailboxMember) =>
+            if (TeamMemberRole.ManagerRole.equals(teamMailboxMember.role.value)) {
+              updateTeamMailboxMembers(teamMailbox, mapMemberNameToRole, mapExistedMembers)
```
chibenwa, line 75: Please flatten the if in the pattern matching. This makes control flow easier to read.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+            } else {
+              SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Not manager of teamMailbox ${teamMailbox.asString()}")))
+            }
+          case None => SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Wrong teamMailboxName ${teamMailbox.asString()}")))
```
chibenwa, line 79: Not found

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+                  mapExistedMembers: Map[Username, TeamMailboxMember]): SMono[TeamMailboxMemberSetResult] = {
+    if (checkIfAnyRolesInvalid(mapMemberNameToRole)) {
+      SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Role could only be manager or member")))
+    } else {
+      if (checkIfAnyManagers(mapMemberNameToRole, mapExistedMembers)) {
+        SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Could not update or remove a manager")))
+      } else {
+        SFlux.fromIterable(mapMemberNameToRole).flatMap(pairMemberNameAndRole => {
+          pairMemberNameAndRole._2 match {
+            case Some(role) => SMono(teamMailboxRepository.addMember(teamMailbox, TeamMailboxMember.of(Username.of(pairMemberNameAndRole._1.value), TeamMemberRole.from(role.role).get)))
```
chibenwa, line 94: Too much indents. Please refactor.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+            case None => SMono(teamMailboxRepository.removeMember(teamMailbox, Username.of(pairMemberNameAndRole._1.value)))
+          }
+        }).collectSeq().`then`(SMono.just(TeamMailboxMemberSetResult(Option[TeamMailboxNameDTO](TeamMailboxNameDTO(teamMailbox.asString())), Option.empty)))
+      }
+    }
+  }
+
+  private def checkIfAnyRolesInvalid(mapMemberNameToRole: Map[TeamMailboxMemberName, Option[TeamMailboxMemberRoleDTO]]): Boolean = {
```
chibenwa, line 102: {} not needed here

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+    mapMemberNameToRole.values.filter(option => option.nonEmpty).map(roleDTO => TeamMemberRole.from(roleDTO.get.role)).exists(maybeRole => maybeRole.isEmpty)
+  }
+
+  private def checkIfAnyManagers(mapMemberNameToRole: Map[TeamMailboxMemberName, Option[TeamMailboxMemberRoleDTO]],
```
chibenwa, line 106: mapMemberNameToRole => memberToRole

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+              updateTeamMailboxMembers(teamMailbox, mapMemberNameToRole, mapExistedMembers)
+            } else {
+              SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Not manager of teamMailbox ${teamMailbox.asString()}")))
+            }
+          case None => SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Wrong teamMailboxName ${teamMailbox.asString()}")))
+      })
+  }
+
+  private def updateTeamMailboxMembers(teamMailbox: TeamMailbox,
+                  mapMemberNameToRole: Map[TeamMailboxMemberName, Option[TeamMailboxMemberRoleDTO]],
```
chibenwa, line 84: mapMemberNameToRole is very unclear to me.

May I suggest to name it after what it represents? 

I suggest membersUpdates.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+                  mapExistedMembers: Map[Username, TeamMailboxMember]): SMono[TeamMailboxMemberSetResult] = {
```
chibenwa, line 85: Idem naming is VERY confusing. Again the strategy to choose what it represent might qualify. I suggest:

`presentMembers`

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private def checkIfAnyManagers(mapMemberNameToRole: Map[TeamMailboxMemberName, Option[TeamMailboxMemberRoleDTO]],
+                                 mapExistedMembers: Map[Username, TeamMailboxMember]): Boolean = {
+    mapExistedMembers.filter(pairMemberNameAndRole => TeamMemberRole.ManagerRole.equals(pairMemberNameAndRole._2.role.value))
+      .exists(pairMemberNameAndRole =>
+        mapMemberNameToRole.keys.map(memberName => Username.of(memberName.value)).toSet.contains(pairMemberNameAndRole._1))
+  }
+
+  private def createTeamMailboxMemberSetResult(teamMailbox: TeamMailbox, setErrorDescription: SetErrorDescription): TeamMailboxMemberSetResult = {
```
chibenwa, line 113: {} not needed

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
chibenwa, line 85: Actually I would even go further: the two types are very similar.
I would even go as far as suggesting writing case class to push the intent to the type system.

```
case class MembersUpdate(value: Map[TeamMailboxMemberName, Option[TeamMailboxMemberRoleDTO]])
case class PresentMembers(value: : Map[Username, TeamMailboxMember])
```

The we could push logic to the case class, example:

```
case class PresentMembers(value: : Map[Username, TeamMailboxMember]) {
  def validate(update: MembersUpdate): Either[SetErrorDescription, MembersUpdate] = {
     // ensure all roles are valid
     // And make sure nobody is manager
     // return the update if valid
   }
}    
```

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  private def updateTeamMailboxMembers(teamMailbox: TeamMailbox,
+                  mapMemberNameToRole: Map[TeamMailboxMemberName, Option[TeamMailboxMemberRoleDTO]],
+                  mapExistedMembers: Map[Username, TeamMailboxMember]): SMono[TeamMailboxMemberSetResult] = {
+    if (checkIfAnyRolesInvalid(mapMemberNameToRole)) {
+      SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Role could only be manager or member")))
+    } else {
+      if (checkIfAnyManagers(mapMemberNameToRole, mapExistedMembers)) {
+        SMono.just(createTeamMailboxMemberSetResult(teamMailbox, SetErrorDescription(s"Could not update or remove a manager")))
+      } else {
```
chibenwa, line 91: Nested if sucks.

Such control flows can be avoided in an idiomatic way in scala by relying on:
 - monads for error management (like `Either`)
 - and using `for ... yield ...` constructs for chaining checks
 
Example:

```
 private def updateTeamMailboxMembers(teamMailbox: TeamMailbox,
     update: MembersUpdate,
     presentMembers: PresentMembers) = {
    val updateResult: Either[SetErrorDescription, SMono[TeamMailboxMemberSetResult]] = for {
       validatedUpdate <- update.validate()
       presentMemersValidateUpdate <- presentMembers.validate(validatedUpdate)
     } yield {
       apply(presentMemersValidateUpdate)
     } 
     updateResult.fold(e =>  SMono.just(createTeamMailboxMemberSetResult(teamMailbox,e))), res => res);
}
```

Clearer?

chibenwa (changes requested):
More time needs to be spent on refactoring code to get something actually reviewable. 
Just getting something working and tested is not enough.
We need to invest in the time to properly describe what is being done (naming, strong typing, method extraction, simplifying control flow, defining the correct abstractions and giving them the correct responsibilities).

FYI I invested 30 minutes of my time on this review.

vttranlina:
Don't forget rebase master

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
hungphan227: what do you mean?

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
hungphan227: I have just made some updates to remove nested if. I guess the code looks good enough and there is no need to apply above stuff.

vttranlina:
a bit my contribute https://github.com/linagora/tmail-backend/commit/ba54428e75f1aab055876181c0dae4183b9220ff
I will pair with @hungphan227 about it

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
hungphan227: updated

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
hungphan227: updated

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/TeamMailboxMemberSetMethodContract.scala
```diff
@@ -0,0 +1,864 @@
...
+           |	"using": ["urn:ietf:params:jmap:core", "urn:ietf:params:jmap:mail", "com:linagora:params:jmap:team:mailboxes"],
+           |	"methodCalls": [
+           |		[
+           |			"TeamMailboxMember/set",
+           |			{
+           |				"accountId": "29883977c13473ae7cb7678ef767cbfbaffc8a44a6e463d971d23a65c1dc4af6",
+           |				"update": {
+           |            "${teamMailbox.asString()}": {
+           |                "${ANDRE.asString()}": null
+           |            }
```
quantranhong1999, line 282: weird indent

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/TeamMailboxMemberSetMethodContract.scala
```diff
@@ -0,0 +1,864 @@
...
+      .body("methodResponses[0]", jsonEquals(
+        s"""[
+           |    "TeamMailboxMember/set",
+           |    {
+           |        "accountId": "29883977c13473ae7cb7678ef767cbfbaffc8a44a6e463d971d23a65c1dc4af6",
+           |        "updated": {},
+           |        "notUpdated": {
+           |            "${teamMailbox.asString()}": {
+           |                "type": "invalidPatch",
+           |                "description": "Team mailbox is not found"
```
quantranhong1999, line 651: Return error message `Team mailbox is not found` when a non-member user tries to update an existing team mailbox sounds misleading to me.

Maybe we could return something like "not manager of the team mailbox`?

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,121 @@
...
+      .map(listResult => TeamMailboxMemberSetResponse.from(request.accountId, listResult))
+
+  private def update(username: Username,
+                     parsingRequestResult: ParsingRequestResult): SMono[TeamMailboxMemberSetResult] =
+    parsingRequestResult match {
+      case ParsingRequestFailure(teamMailboxNameDTO, exception) => SMono.just(TeamMailboxMemberSetResult.notUpdated(teamMailboxNameDTO, SetErrorDescription(exception.message)))
+      case ParsingRequestSuccess(teamMailbox, membersUpdateToAdd, membersUpdateToRemove) =>
+        checkIfAnyUsersDoesNotExist(membersUpdateToAdd, membersUpdateToRemove)
+          .filter(anyUsersNotExist => anyUsersNotExist)
+          .map(_ => TeamMailboxMemberSetResult.notUpdated(teamMailbox, SetErrorDescription("Some users do not exist in the system")))
```
quantranhong1999, line 63: Listing which users do not exist would be better IMO.

Imagine updating multiple users at the same request. It is hard to know which user has the existance issue.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/model/TeamMailboxMemberSetRequest.scala
```diff
@@ -0,0 +1,100 @@
...
+import cats.implicits._
+import com.linagora.tmail.team.{TeamMailbox, TeamMailboxMember}
+import org.apache.james.core.Username
+import org.apache.james.jmap.core.SetError.SetErrorDescription
+import org.apache.james.jmap.core.{AccountId, SetError}
+import org.apache.james.jmap.method.WithAccountId
+
+import scala.util.Try
+
+trait TMBParseException extends RuntimeException {
```
quantranhong1999, line 12: Not using acronym please

quantranhong1999 (commented):
Looks good overall to me.

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/TeamMailboxMemberSetMethodContract.scala
hungphan227: fixed

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,121 @@
...
+      .`then`(SMono.just(TeamMailboxMemberSetResult(Option[TeamMailboxNameDTO](TeamMailboxNameDTO(teamMailbox.asString())), Option.empty)))
+
+  private def addMember(teamMailbox: TeamMailbox, members: List[TeamMailboxMember]): SMono[Unit] =
+    SFlux.fromIterable(members)
+      .flatMap(member => teamMailboxRepository.addMember(teamMailbox, member))
+      .`then`()
+
+  private def removeMember(teamMailbox: TeamMailbox, usernames: Set[Username]): SMono[Unit] =
+    SFlux.fromIterable(usernames)
+      .flatMap(username => teamMailboxRepository.removeMember(teamMailbox, username))
```
chibenwa, line 119: Please control concurrency on all Flux / SFlux flatMaps

chibenwa, line 119: Feeling sad that we actually end up splitting our updates into individual inserts / remove.
Appart from generating more requests we loose atomicity of the requests.
That's sad because the underlying layers (mailbox ACL) do support full/partial updates in an atomic fashion.
A well designed reposiotry layer would have had exposed this inner structure and allowed to leverage atomicity onto the presentation layer.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,121 @@
...
+          }
+        case None => SMono.just(TeamMailboxMemberSetResult.notUpdated(teamMailbox, SetErrorDescription("Team mailbox is not found or not a member of the mailbox")))
+      })
+
+  private def updateTeamMailboxMembers(teamMailbox: TeamMailbox,
+                                       membersUpdateToAdd: List[TeamMailboxMember],
+                                       membersUpdateToRemove: Set[Username],
+                                       presentMembers: Map[Username, TeamMailboxMember]): SMono[TeamMailboxMemberSetResult] =
+    SFlux.fromIterable(membersUpdateToRemove.concat(membersUpdateToAdd.map(teamMailboxMember => teamMailboxMember.username)))
+      .filter(username => presentMembers.contains(username) && ManagerRole.equals(presentMembers(username).role.value))
```
chibenwa, line 100: We can avoid the complex boolean condition with an Option + default clause.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,121 @@
...
+          .map(usernames => TeamMailboxMemberSetResult.notUpdated(teamMailbox,
+            SetErrorDescription(s"Some users do not exist in the system: ${usernames.map(username => username.asString()).toArray.mkString("", ", ", "")}")))
+          .switchIfEmpty(updateTeamMailboxMembers(username, teamMailbox, membersUpdateToAdd, membersUpdateToRemove))
+          .onErrorResume {
+            case _: TeamMailboxNotFoundException => SMono.just(TeamMailboxMemberSetResult.notUpdated(teamMailbox, SetErrorDescription("Team mailbox is not found or not a member of the mailbox")))
+            case _: Throwable => SMono.just(TeamMailboxMemberSetResult.notUpdated(teamMailbox, SetErrorDescription("Internal error")))
+          }
+    }
+
+  private def checkIfAnyUsersDoesNotExist(membersUpdateToAdd: List[TeamMailboxMember],
```
chibenwa, line 73: I spent 5 minutes just understanding this does listNonExisting users.

Please use good names.

chibenwa (commented):
I find it really tricky to read this piece of code as the asynchronous control flow is mixed with data validation and checks are scattered all other the place.

chibenwa:
We need audit trail for team mailbox member modifications. Who does it and who get access, loose access.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/model/TeamMailboxMemberSetRequest.scala
```diff
@@ -0,0 +1,100 @@
...
+
+case class InvalidRoleException(role: TeamMailboxMemberRoleDTO) extends ParsingRequestException {
+  override def message: String = s"Invalid role: ${role.role}"
+}
+
+case class InvalidTeamMemberNameException(memberName: TeamMailboxMemberName) extends ParsingRequestException {
+  override def message: String = s"Invalid team member name: ${memberName.value}"
+}
+
+trait ParsingRequestResult
```
vttranlina, line 29: Should better if has prefix name
eg: `TeamMailboxMemberParsingRequestResult`

or relocate it to child package
eg: `com.linagora.tmail.james.jmap.model.teammailbox.member.ParsingRequestResult`

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,121 @@
...
+                                membersUpdateToRemove: Set[Username]): SFlux[Username] =
```
vttranlina, line 74: indent
Screenshot_20240705_171903

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
vttranlina: and return `SMono[Seq[Username]]` 
Then, at line 61 it will be pretty

```scala
listNonExistingUsers(membersUpdateToAdd, membersUpdateToRemove)
  .flatMap {
    case Seq.empty => updateTeamMailboxMembers(username, teamMailbox, membersUpdateToAdd, membersUpdateToRemove)
    case unExistUsernames => SMono.just(TeamMailboxMemberSetResult.notUpdated(teamMailbox,
      SetErrorDescription(s"Some users do not exist in the system: ${unExistUsernames.map(username => username.asString()).toArray.mkString("", ", ", "")}")))
  }
}
```

chibenwa:
https://github.com/chibenwa/openpaas-james/commit/e4caf1cb80b95e34eca52d447e0649980b94a0eb

This is the kind of quality I expect from team members in terms of code organisation.

BTW it took me over an hour to put this together.

Arsnael:
Read it nothing to add

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
hungphan227: Move this issue to another ticket https://github.com/linagora/james-project/issues/5221

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,138 @@
...
+      .map(listResult => TeamMailboxMemberSetResponse.from(request.accountId, listResult))
+
+  private def performUpdate(username: Username, parsingRequestResult: TeamMailboxMemberParsingRequestResult): SMono[TeamMailboxMemberSetResult] =
+  {
+    val r: SMono[TeamMailboxMemberSetResult] = for {
+      requestWithValidSyntax <- validateSyntax(parsingRequestResult)
+      requestWithUserExistenceValidation <- validateUsersExist(requestWithValidSyntax)
+      presentMembers <- retrievePresentMembers(requestWithUserExistenceValidation)
+      requestWithAuthorizationValidation <- validateAuthorizations(username, requestWithUserExistenceValidation, presentMembers)
+      requestWithAuthorizationValidation <- validateNoChangeToManagers(requestWithAuthorizationValidation, presentMembers)
```
vttranlina, line 64: a bit confuse,  when see double variable has same name `requestWithAuthorizationValidation` 
it will better:
`requestWithAuthorizationValidation` -> `validatedRequest`

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
```diff
@@ -0,0 +1,138 @@
...
+      case e: TeamMailboxNotFoundException => SMono.just(TeamMailboxMemberSetResult.notUpdated(e.teamMailbox, SetErrorDescription("Team mailbox is not found or not a member of the mailbox")))
+      case e: ManagerAuthorizationChanged => SMono.just(TeamMailboxMemberSetResult.notUpdated(e.teamMailbox, SetErrorDescription(s"Could not update or remove manager ${e.username.asString()}")))
+      case e: InvalidSyntax => SMono.just(TeamMailboxMemberSetResult.notUpdated(e.failure.teamMailboxName, SetErrorDescription(e.failure.exception.message)))
+    }
+  }
+
+  private def retrievePresentMembers(requestWithUserExistenceValidation: ParsingRequestSuccess): SMono[Map[Username, TeamMailboxMember]] =
+    SFlux(teamMailboxRepository.listMembers(requestWithUserExistenceValidation.teamMailbox)).collectMap[Username, TeamMailboxMember](member => member.username, member => member)
+
+  private case class NonExistingUsers(teamMailbox: TeamMailbox, usersNotFound: Seq[Username]) extends RuntimeException
```
vttranlina, line 82: Benoit's code is just a draft. It's not a big convention deal, but we can create an object TeamMailboxMemberSetMethod and move the exception case class into it.

vttranlina (approved):

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/TeamMailboxMemberSetMethod.scala
chibenwa: From what I could read you are about to attempt a massive reset of all the rights, which is not idempotent and even a regression.

Maybe we cannot achieve full atomicity.

I thought the ACL command was multi-valued (one command -> several keys)

It is actually not multi-valued so existing mechanisms do not add benefits.

Sorry for bothering.

Arsnael (commented):
Read it nothing to add. Almost there!

Arsnael (approved):

quantranhong1999 (approved):

chibenwa (approved):

PR #1147, opened 2024-07-30
Upgrade apisix base image to 3.9.1-debian

Arsnael:
From the local docker compose test I could check quickly it seems to still work. But let's test it first on a staging environment before moving to it

### DIFF
--- a/demo/apisix/Dockerfile
+++ b/demo/apisix/Dockerfile
@@ -1,4 +1,4 @@
-FROM apache/apisix:3.2.0-debian
+FROM apache/apisix:3.9.1-debian
 USER root
 RUN apt-get update && apt -y install openjdk-11-jre-headless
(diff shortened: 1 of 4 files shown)


quantranhong1999 (approved):

chibenwa (approved):

vttranlina (commented):
3.2.2 is an LTS version.
3.9.1 is not an LTS version. Is it a big deal?

Arsnael:
First, it's 3.2.0 used here, not 3.2.2 :)

Then seems so on their website that 3.2.2 they call it lts. However I can see that 3.2.2 was released... a year ago? Even the 3.2.2 docker images on the apisix docker repo are dated form a year ago... is it really a LTS with continuous security updates and such? I doubt it^^'

vttranlina (approved):

PR #1151, opened 2024-08-06
[FIX] tmail-backend distributed should be able to run without Redis

quantranhong1999:
The `redis.properties` was mounted by default. The existence of the file would force tmail-backend to load Redis modules as of today and fail startup when no Redis is bound.

Mounting the file optional would allow Redis usage to be optional.


tmail-backend/apps/distributed/docker-compose.yml
```diff
@@ -36,6 +36,7 @@ services:
     volumes:
       - ${RSA_PUBLICKEY_PATH}:/root/conf/jwt_publickey # Replace with absolute path to your RSA public key
       - ${RSA_PRIVATEKEY_PATH}:/root/conf/jwt_privatekey # Replace with absolute path to your RSA private key
+      - ./redis.properties:/root/conf/redis.properties # Replace with absolute path to your RSA private key
```
Arsnael, line 39: Should put in an other folder I think instead of root folder, like we do for some apps on james, like docker-configuration?

Arsnael (approved):

PR #1173, opened 2024-09-06
Distributed app: default config to relax mailbox name validation

quantranhong1999:


Arsnael (commented):
Do we really need to enable it by default on tmail?

quantranhong1999:
> Do we really need to enable it by default on tmail?

Answered in: https://github.com/linagora/tmail-backend/pull/1174#issuecomment-2333758581

But this could just be optional for the ticket though.

It is weird that `DistributedImapAndSmtpTest` failed on CI on both PRs with opensearch container failed startup error.
I just tested locally and it is fine...

Arsnael:
I saw the opensearch been failing regularly to start lately on tmail-backend... Also seems we use an image for those depoyment tests that's a bit outdated (2.1.0). Can we match the one used in james? Maybe the image is more stable? (hard gamble here I'm not really convinced but...)

quantranhong1999:
> Can we match the one used in james? Maybe the newer image is more stable?

https://github.com/linagora/tmail-backend/pull/1176

Arsnael (approved):

PR #1174, opened 2024-09-06
Postgres app: default config to relax mailbox name validation

quantranhong1999:


Arsnael (commented):
Do we really need to enable it by default on tmail?

quantranhong1999:
> Do we really need to enable it by default on tmail?

Previously Benoit had some issue migrating user mailboxes. So I think it is good to make it default in our default conf.

Arsnael:
The warning mention in the doc would make me think we should probably not enable it by default (and if there is a need just enable it)

@chibenwa ?

PR #1178, opened 2024-09-09
TMail: unset JAVA_TOOL_OPTIONS env variable when running james-cli

quantranhong1999:
GIVEN I run james-cli in kubernetes
THEN the pod crashes and reboots

Because the james-cli runs JVM it catches the environment variables set by James. As such it always pretouch 3GB of RAM. When added to memory consumed by James it exceeds the limit and pod get's OOM-killed.

### DIFF
--- a/tmail-backend/apps/distributed/src/main/scripts/james-cli
+++ b/tmail-backend/apps/distributed/src/main/scripts/james-cli
@@ -1,3 +1,4 @@
 #!/bin/bash
 
+unset JAVA_TOOL_OPTIONS
 java -cp /root/resources:/root/classes:/root/libs/* org.apache.james.cli.ServerCmd "$@"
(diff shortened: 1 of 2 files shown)


Arsnael (approved):

chibenwa (approved):

Arsnael:
Openai mailet issue again. You might need to match with your fix on master branch @quantranhong1999 :)

quantranhong1999:
> Openai mailet issue again. You might need to match with your fix on master branch @quantranhong1999 :)

Just curious but is the build supposed to be rebased on the latest master (with the openai fix)? Anyway I would do a rebase for sure.

Arsnael:
Mixed up PRs, thought was postgres branch apologies.

Normally yes but then do a rebase yes, maybe better in this case

Issue #1109, opened 2024-06-26
Mail GPT: feed user context


chibenwa:
### Why?

Import mail knowlege into my personnal GPT context thus improving relevance.

### What?

 - Configuration (of the extensions) for API creds and endpoints of the GPT trainer
 - Configuration (on/off, off by default) of feeding data to the GPT model via JMAP settings, making it eventually available as user setting.
 - Mailbox listener feeding the model for people activating the option

chibenwa:
Idea: store context onto Redis?

HoussemNasri:
Idea: Display a list of common prompts to the user.

Examples: 
1. Summarize unread emails from the past week.
2. Sort emails by priority based on content.
3. etc

Arsnael:
Moving this to In Progress column, as @HoussemNasri  is wroking on it actually

chibenwa:
Duplicates #1795

PR #1196, opened 2024-09-23
ISSUE-1109 Integrate ai.linagora.com for automated AIBot response generation

HoussemNasri:
# Summary of changes
- Renamed OpenAIMailet to AIBotMailet.
- Added MailBotConfig to save the mailet from the responsibility of configuration handling.
- Added support for using LLM models from ai.linagora.com.
- Changed the format of the <model/> parameter in the MainBot config to include the LLM in addition to the model: {llm},{llm_model}
- Added abstractions to prepare for the addition of more models.

## TODO

- ~~[ ] Rename the `tmail-open-ai` maven module.~~ Since we are only supporting OpenAI compatible APIs the name can be kept the same.
- [x] Update documentation.
- [ ] Add integration tests for ai.linagora.com.


demo/tmail/mailetcontainer.xml
```diff
@@ -98,9 +98,10 @@
             </mailet>
 
             <!-- Put the OpenAIMailet after LocalDelivery so the GPT reply would come after the asking question -->
-            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.OpenAIMailet">
+            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.MailBotMailet">
```
chibenwa, line 101:
```suggestion
            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.AIBotMailet">
```

?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/ChatLanguageModelFactory.java
```diff
@@ -0,0 +1,27 @@
...
+    private static final String OPEN_AI_BASE_URL = "https://api.openai.com/v1";
+    private static final String LINAGORA_AI_BASE_URL = "https://ai.linagora.com/api/";
+
+    public ChatLanguageModel createChatLanguageModel(MailBotConfig config) {
+        String apiKey = config.getApiKey();
+        LlmModel llmModel = config.getLlmModel();
+
+        return switch (llmModel.llm()) {
+            case OPEN_AI -> createOpenAILanguageModel(apiKey, llmModel.modelName(), OPEN_AI_BASE_URL);
+            case LINAGORA_AI -> createOpenAILanguageModel(apiKey, llmModel.modelName(), LINAGORA_AI_BASE_URL);
```
chibenwa, line 16: To be faily honnest I'd prefer having a `<baseURL>` property in the mailet configuration and not hard code ai.linagra.com URL in the code.

quantranhong1999 (commented):
Read it. Looks good so far to me.

demo/tmail/mailetcontainer.xml
HoussemNasri: Okay no problem, but I would have to rename it everywhere in a separate commit.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/ChatLanguageModelFactory.java
HoussemNasri: Sure, it shouldn't be hard to add. That would allow users to submit their custom OpenAI-based models also, I like the idea.

HoussemNasri: Done. Now, the Llm enum refers to the API type rather than a specific LLM. I'm thinking of renaming it to LlmAPI or LlmType, any suggestions of better names.

HoussemNasri:
Question: I want to add integration tests for the Linagora AI API to ensure that after setting the base URL to https://ai.linagora.com/api, the API calls are successful. However, I'm unsure how to securely use an API key in these tests. Hardcoding it seems unwise. What’s the best approach to handle this?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -0,0 +1,114 @@
...
+    public static String API_KEY_PARAMETER_NAME = "apiKey";
+    public static String GPT_ADDRESS_PARAMETER_NAME = "gptAddress";
+    public static String MODEL_PARAMETER_NAME = "model";
+    public static String BASE_URL_PARAMETER_NAME = "baseURL";
+
+    public static final LlmModel DEFAULT_LLM_MODEL =
+            new LlmModel(Llm.OPEN_AI, "gpt-4o-mini");
+
+    private final String apiKey;
+    private final String baseURL;
```
chibenwa, line 25: Can we strongly type this here?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (AddressException e) {
+            throw new RuntimeException(e);
+        }
+
+        LlmModel llmModel;
+        if (Strings.isNullOrEmpty(llmModelParam)) {
+            llmModel = DEFAULT_LLM_MODEL;
+        } else {
+            llmModel = parseLlmModelParamOrThrow(llmModelParam);
+        }
```
chibenwa, line 66: This variable reallocation can eaasily be avoided with variable extraction.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -0,0 +1,114 @@
...
+        if (Strings.isNullOrEmpty(gptAddressParam)) {
+            throw new MailetException("No value for " + GPT_ADDRESS_PARAMETER_NAME + " parameter was provided.");
+        }
+        
+        MailAddress mailAddress;
+        try {
+            mailAddress = new MailAddress(gptAddressParam);
+        } catch (AddressException e) {
+            throw new RuntimeException(e);
+        }
```
chibenwa, line 59: Can we widen the exception thrown to avoid variable reallocation?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -0,0 +1,114 @@
...
+        String[] tokens = llmModelParam.split(",");
+
+        if (tokens.length != 2) {
+            throw new MailetException("""
+                    Invalid LLM model parameter. Please add the LLM along with the specific model to use.
+                    Example: openai,gpt-4-mini
+                    Provided value: %s""".formatted(llmModelParam));
+        }
+
+        String llm = tokens[0];
```
chibenwa, line 81: Not used as far as I can tell except in the exception. Please drop the llm part of the model.

HoussemNasri: It's used in `ChatLanguageModelFactory`, we need it to decide what implementation of ChatLanguageModel to use. For now, we only use OpenAIChatLangaugeModel, in the future more implementation would be used.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: Why is variable reallocation bad?

HoussemNasri:
> Question: I want to add integration tests for the Linagora AI API to ensure that after setting the base URL to https://ai.linagora.com/api, the API calls are successful. However, I'm unsure how to securely use an API key in these tests. Hardcoding it seems unwise. What’s the best approach to handle this?

One option is to follow `FirebaseClientTest`:
1. Read the API key from the environment using `System.getenv()`.
2. Disable the test class by default to indicate that it should be run manually, preventing the need for additional configuration.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
florentos17: i think it's just for code readability, it's better if a variable always refers to the same thing :)

Arsnael:
Hi @HoussemNasri .

Are you working on this? https://github.com/linagora/tmail-backend/issues/1109

If yes please then assign the task to yourself and switch the status to in progress.

Also you should add the link of this PR in the ticket, and precede your commits description with the issue number, like 

`ISSUE-1109 [description]`

Thanks

tmail-backend/tmail-third-party/open-ai/AIBotMailet.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+== Configuration Example
+
+[source,xml]
+----
+<mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@linagora.com" class="com.linagora.tmail.mailet.AIBotMailet">
+    <apiKey>demo</apiKey>
+    <gptAddress>gpt@tmail.com</gptAddress>
+    <model>openai,gpt-4o-mini</model>
+</mailet>
```
Arsnael, line 13: baseUrl missing in the example

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotMailet.java
```diff
@@ -88,35 +84,35 @@ public void service(Mail mail) {
...
+    private String askChatLanguageModel(Mail mail) throws IOException, MessagingException {
         ChatMessage systemMessage = new SystemMessage("You are a helpful mailing chatbot. Respond to the following email as a chatbot, not as the recipient. ");
         ChatMessage userMessage = new UserMessage(evaluatePrompt(mail));
 
-        return openAiModel.generate(systemMessage, userMessage)
-            .content()
-            .text();
+        return chatLanguageModel.generate(systemMessage, userMessage)
+                .content()
+                .text();
```
Arsnael, line 107: unnecessary indent

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotMailet.java
```diff
@@ -88,35 +84,35 @@ public void service(Mail mail) {
...
     }
 
     private String evaluatePrompt(Mail mail) throws IOException, MessagingException {
         MessageContentExtractor.MessageContent askingMessageContent = extractMessageContent(mail);
         Optional<String> content = askingMessageContent.extractMainTextContent(htmlTextExtractor);
 
         return Strings.nullToEmpty(mail.getMessage().getSubject()) + "\n" +
-            content.orElse("");
+               content.orElse("");
```
Arsnael, line 115: idem

Arsnael: and for the rest too, maybe check your IDE settings?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/ChatLanguageModelFactory.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class ChatLanguageModelFactory {
+
+    public ChatLanguageModel createChatLanguageModel(AIBotConfig config) {
+        String apiKey = config.getApiKey();
+        LlmModel llmModel = config.getLlmModel();
+        Optional<URL> baseURLOpt = config.getBaseURL();
+
+        return switch (llmModel.llm()) {
+            case OPEN_AI -> createOpenAILanguageModel(apiKey, llmModel.modelName(), baseURLOpt.map(URL::toString).orElse(null));
```
Arsnael, line 17: I'm missing something here. Can you tell me what happens in case of having baseURL empty? Where do we get the endpoint used for AI API?

Also I would advocate the usage of null if possible

Arsnael:
https://james-jenkins.lin-saas.com/job/Tmail%20build/job/PR-1196/5/testReport/junit/com.linagora.tmail.mailet/AIBotMailetTest/shouldReplyToSender/

Perfect example of why you should avoid null :)

tmail-backend/tmail-third-party/open-ai/AIBotMailet.adoc
HoussemNasri: The need for a base URL is the special case here, normally, langchain4j (the library we use to interact with the different LLMs APIs) would know which base URL to use by default, for the different LLMs it supports (OpenAI, Mistral, Llama, etc).

However, the library does not support ai.linagora.com. Fortunately, AI Liangora is compatible with the OpenAPI API, which enables us to use the OpenAI API implementation in langchain4j and change the baseURL.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotMailet.java
HoussemNasri: Please see comment above.

HoussemNasri:
> Perfect example of why you should avoid null :)

That is true, but I believe keeping baseURL optional is the right decision here. The exception is thrown because of a bug in the implementation, null checks should be added. We only have to worry about nullability when parsing the configurations. The `getURL` method in `AIMailConfig` returns an `Optional`.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/ChatLanguageModelFactory.java
HoussemNasri: Please see the comment above.

Arsnael:
> That is true, but I believe keeping baseURL optional is the right decision here. The exception is thrown because of a bug in the implementation, null checks should be added. We only have to worry about nullability when parsing the configurations. The `getBaseURL` method in `AIMailConfig` returns an `Optional`.

You can easily keep your logic around the optional as well without the null?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
chibenwa: Yes here it is not that hard but it can quickly go out of control.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/Llm.java
```diff
@@ -0,0 +1,28 @@
+package com.linagora.tmail.mailet;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+public enum Llm {
+    OPEN_AI("openai");
```
chibenwa, line 9: We only support openai as of today.

Let's not over anticipate future need, remove this constant and related configuration?

demo/tmail/mailetcontainer.xml
```diff
@@ -98,9 +98,10 @@
             </mailet>
 
             <!-- Put the OpenAIMailet after LocalDelivery so the GPT reply would come after the asking question -->
-            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.OpenAIMailet">
+            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.AIBotMailet">
                 <apiKey>demo</apiKey>
                 <gptAddress>gpt@tmail.com</gptAddress>
+                <model>openai,gpt-4o-mini</model>
```
chibenwa, line 104: We have configuration to structure things, why parsing a configuration field?

I'd prefer :

```
            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.AIBotMailet">
                <apiKey>demo</apiKey>
                <gptAddress>gpt@tmail.com</gptAddress>
                <vendor>openai</vendor>
                <model>gpt-4o-mini</model>
            </mailet>
        </processor>
```

We should drop BTW the vendor part as only one implem is available today.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: Benoit can you explain in more details what you mean by avoid variable reallocation, what code changes can I make?

chibenwa: Read the book "Clean code"

https://community.sonarsource.com/t/new-rule-variable-should-not-be-reassigned/78154

Also on communication avoid designating a single individual. Aren't other able to provide the explaination?

chibenwa: (Also another great tip is to avoid 'you' and favor 'we' - this forces to think collective together and prevent antagonising people)

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -37,7 +35,7 @@ public class AIBotConfig {
     private final MailAddress gptAddress;
     private final LlmModel llmModel;
 
-    public AIBotConfig(String apiKey, @Nullable URL baseURL, MailAddress gptAddress, LlmModel llmModel) {
+    public AIBotConfig(String apiKey, MailAddress gptAddress, LlmModel llmModel, @Nullable URL baseURL) {
```
chibenwa, line 38: Why @Nullable? 

Can we please avoid using null and rely on a strong type (ie a monad) like `Optional` to represent the absence of something?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -37,7 +35,7 @@ public class AIBotConfig {
...
         Objects.requireNonNull(apiKey);
         Objects.requireNonNull(gptAddress);
         Objects.requireNonNull(llmModel);
```
chibenwa, line 41: 1. Preconditions.checkNotNull to be consistant with team practices

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -61,31 +59,30 @@ public static AIBotConfig fromMailetConfig(MailetConfig mailetConfig) throws Mai
...
         }
-        
-        MailAddress mailAddress;
+
+        Optional<URL> baseURLOpt = toOptionalIfNotEmpty(baseUrlParam)
+                .flatMap(AIBotConfig::baseURLStringToURL);
+
+        LlmModel llmModel = !Strings.isNullOrEmpty(llmModelParam)
+                            ? parseLlmModelParamOrThrow(llmModelParam)
+                            : DEFAULT_LLM_MODEL;
```
chibenwa, line 68: Use optionals in order to avoid conditions?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
chibenwa, line 38: I am unhappy that we take decisions in this constructor (default values). Maybe calling code can resolve the default value and just pass non nullable values here?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -131,4 +128,12 @@ public Optional<URL> getBaseURL() {
             return Optional.of(baseURL);
         }
     }
+
+    private static Optional<String> toOptionalIfNotEmpty(String s) {
+        if (Strings.isNullOrEmpty(s)) {
```
chibenwa, line 133: Optional.ofNullable(...).filter(...) ?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -61,31 +59,30 @@ public static AIBotConfig fromMailetConfig(MailetConfig mailetConfig) throws Mai
...
-        LlmModel llmModel;
-        if (Strings.isNullOrEmpty(llmModelParam)) {
-            llmModel = DEFAULT_LLM_MODEL;
-        } else {
-            llmModel = parseLlmModelParamOrThrow(llmModelParam);
-        }
+    }
 
-        URL baseURL = null;
+    private static @NotNull Optional<URL> baseURLStringToURL(String baseUrlString) {
```
chibenwa, line 78: We generaly do not use @NotNull annotations in this code base

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -61,31 +59,30 @@ public static AIBotConfig fromMailetConfig(MailetConfig mailetConfig) throws Mai
...
-            if (!Strings.isNullOrEmpty(baseUrlParam)) {
-                baseURL = URI.create(baseUrlParam).toURL();
-            }
+            return Optional.of(URI.create(baseUrlString).toURL());
         } catch (MalformedURLException e) {
             LOGGER.warn("Invalid LLM API base URL", e);
         }
 
-        return new AIBotConfig(apiKeyParam, baseURL, mailAddress, llmModel);
+        return Optional.empty();
```
chibenwa, line 85: We are accepting invalid URL, ignoring them and just start with the default one.

Good chance the user do not notice easily what's wrong.

Aborting the configuration sequence in case of bad values is perfectly Ok.

IMO we should propagate this exception.

tmail-backend/tmail-third-party/open-ai/AIBotMailet.adoc
Arsnael: Fair, thanks for the explanation. Shouldn't that be precised in the doc though then?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: That would mean we need to pass an Optional as an argument? I did that but it gets highlighted in yellow in IntelliJ saying it is against best practices.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: I wasn't aware of the `Optional.ofNullable(...).filter(...)` trick, that's really helpful!

tmail-backend/tmail-third-party/open-ai/AIBotMailet.adoc
HoussemNasri: Makes sense, documentation updated.

demo/tmail/mailetcontainer.xml
HoussemNasri: Done.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: That is exactly what is happening, `fromMailetConfig` calls the constructor with valid values, the assertions were added just in case.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: Done.

demo/tmail/mailetcontainer.xml
HoussemNasri: The idea here was to simplify default resolution: what if the user provided a model but not a vendor or the reverse.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/ChatLanguageModelFactory.java
HoussemNasri: Not relevant anymore as the LLM enum is deleted.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
chibenwa: Then why do we have a @Nullable parameter?

demo/tmail/mailetcontainer.xml
chibenwa, line 104:
```suggestion
                <model>gpt-4o-mini</model>
```

I insist

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -0,0 +1,106 @@
...
+    public LlmModel getLlmModel() {
+        return llmModel;
+    }
+
+    public Optional<URL> getBaseURL() {
+        if (baseURL == null) {
+            return Optional.empty();
+        } else {
+            return Optional.of(baseURL);
+        }
```
chibenwa, line 104: Optional.offNullable

BTW why isn't baseUrl an optional itself?

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
```diff
@@ -0,0 +1,106 @@
...
+            throw new RuntimeException(e);
+        }
+
+    }
+
+    private static Optional<URL> baseURLStringToURL(String baseUrlString) {
+        try {
+            return Optional.of(URI.create(baseUrlString).toURL());
+        } catch (MalformedURLException e) {
+            LOGGER.warn("Invalid LLM API base URL", e);
```
chibenwa, line 81: Forward the exception please

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
chibenwa: > That would mean we need to pass an Optional as an argument? 

Better than `null`

>  I did that but it gets highlighted in yellow in IntelliJ saying it is against best practices.

I know Intellij tend to do warning about this. Alternatives might be to create more subtypes.

Yet to me, having something absent is perfectly fine, and handling it through the type system is 1000x better.

Please disregard this IDE warning (unless if we can come up with a cleaner deasign)

HoussemNasri: Don't think we can get a cleaner design, passing an Optional then.

HoussemNasri: Made it Optional.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: It is now.

tmail-backend/tmail-third-party/open-ai/src/main/java/com/linagora/tmail/mailet/AIBotConfig.java
HoussemNasri: Done.

tmail-backend/tmail-third-party/open-ai/AIBotMailet.adoc
```diff
@@ -0,0 +1,24 @@
...
+      <baseURL>https://ai.linagora.com/api</baseURL>
+</mailet>
+----
+
+=== Parameters
+
+The mailet configuration requires the following parameters:
+
+* `apiKey`: The API key for accessing the LLM API service.
+* `gptAddress`: The email address used to send replies generated by the LLM model.
```
Arsnael, line 22: Now that I think about it again (sorry to be picky^^): as we are now making a more generic API that could use any LLM languages and not just chatGPT, does that parameter name make sense?

Maybe something like `botAddress` would make more sense for example?

HoussemNasri: You are right, the parameter doesn't make sense anymore, it should be good now.

chibenwa (approved):

Arsnael (approved):

PR #1205, opened 2024-10-01
ISSUE-1204 Autocomplete prefix search for names

quantranhong1999:
Previously, prefix search for first name/surname works but only with at least 3 chars. Reason: to not return too noise suggestions, and save indexing space.

However, Diana seems to have requested this to be able to search with fewer characters. Verified with GMail that supports autocomplete addresses from 1 char though.

TODO: testing and document index (settings) migration steps.

### DIFF
--- a/tmail-backend/jmap/extensions-opensearch/src/main/java/com/linagora/tmail/james/jmap/OpenSearchContactConfiguration.java
+++ b/tmail-backend/jmap/extensions-opensearch/src/main/java/com/linagora/tmail/james/jmap/OpenSearchContactConfiguration.java
@@ -104,7 +104,7 @@ public class OpenSearchContactConfiguration {
     public static final WriteAliasName DEFAULT_ALIAS_WRITE_DOMAIN_CONTACT_NAME = new WriteAliasName("domain_contact_write_alias");
     public static final ReadAliasName DEFAULT_ALIAS_READ_DOMAIN_CONTACT_NAME = new ReadAliasName("domain_contact_read_alias");
     public static final Integer DEFAULT_MAX_NGRAM_DIFF = 27;
-    public static final Integer DEFAULT_MIN_NGRAM = 3;
+    public static final Integer DEFAULT_MIN_NGRAM = 1;
 
     public static final OpenSearchContactConfiguration DEFAULT_CONFIGURATION = builder().build();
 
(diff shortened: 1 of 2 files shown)


tmail-backend/jmap/extensions-opensearch/src/main/java/com/linagora/tmail/james/jmap/OpenSearchContactConfiguration.java
```diff
@@ -104,7 +104,7 @@ public static Builder builder() {
     public static final WriteAliasName DEFAULT_ALIAS_WRITE_DOMAIN_CONTACT_NAME = new WriteAliasName("domain_contact_write_alias");
     public static final ReadAliasName DEFAULT_ALIAS_READ_DOMAIN_CONTACT_NAME = new ReadAliasName("domain_contact_read_alias");
     public static final Integer DEFAULT_MAX_NGRAM_DIFF = 27;
-    public static final Integer DEFAULT_MIN_NGRAM = 3;
+    public static final Integer DEFAULT_MIN_NGRAM = 1;
```
chibenwa, line 107: 2 as a minimum please. We are not google.

chibenwa (commented):
Otherwise +1

tmail-backend/jmap/extensions-opensearch/src/main/java/com/linagora/tmail/james/jmap/OpenSearchContactConfiguration.java
HoussemNasri: Why 2 is better?

tmail-backend/jmap/extensions/src/test/scala/com/linagora/tmail/james/jmap/contact/EmailAddressContactSearchEngineContract.scala
```diff
@@ -52,6 +54,28 @@ trait EmailAddressContactSearchEngineContract {
 
   def awaitDocumentsIndexed(query: QueryType, documentCount: Long): Unit
 
+  @ParameterizedTest
+  @ValueSource(strings = Array("d", "di", "dia", "dian", "diana"))
```
HoussemNasri, line 58: "d" should be removed now that the minimum is 2?

quantranhong1999: Yes, good catch

tmail-backend/jmap/extensions-opensearch/src/main/java/com/linagora/tmail/james/jmap/OpenSearchContactConfiguration.java
quantranhong1999: > Why 2 is better?

I explained in the commit message: `Reason: to not return too noise suggestions, and save indexing space.`

HoussemNasri: Thanks!

chibenwa (approved):

HoussemNasri (approved):
LGTM!

PR #1206, opened 2024-10-02
ISSUE-1204 Documentation contact indices settings migration

quantranhong1999:


upgrade-instructions.md
```diff
@@ -12,23 +12,325 @@ software documentation. Do not follow this guide blindly!
...
+      },
+      "email": {
+        "type": "text",
+        "analyzer": "email_ngram_filter_analyzer",
+        "search_analyzer": "rebuilt_keyword"
+      },
+      "firstname": {
+        "type": "text",
+        "analyzer": "name_edge_ngram_filter_analyzer",
+        "search_analyzer": "standard"
```
chibenwa, line 104: Is this intended to keep standard here?

quantranhong1999: I think so. What is the expectation to not keep search_analyzer as standard?

chibenwa: Having a `standard` search analyzer means filtering.

Example

GIVEN `Andre` 
WHEN I type `and` in the search bar
THEN the standard analyzer thinks it is an english stopword and filters out the token thus rendering the search useless.

Treating the search as a keyword (tokenized the right way) would avoid such pitfalls.

I'm interested by the test of andre/and BTW to see if I am right.

HoussemNasri: Relevant documentation: https://opensearch.org/docs/latest/analyzers/

HoussemNasri: `Keyword` seem like the analyzer we want to use here, given that we don't want to split the emails before indexing.

quantranhong1999: > I'm interested by the test of andre/and BTW to see if I am right.

I added a test. We should be good already.

quantranhong1999: > Overall Iam sory I was convinced the standard analyzer was doing more work here.

Don't be sorry. It is always good to check though.

Arsnael (approved):

chibenwa (approved):

PR #1212, opened 2024-10-03
ISSUE-1208 Allow use of custom schema for LDAP Mailing lists

quantranhong1999:
`businessCategory: cn=memberRestrictedList,ou=twakeListType,ou=nomenclature,o=gov,c=mu`

=> businessCategory = memberRestrictedList

### DIFF
--- a/tmail-backend/mailets/src/main/java/com/linagora/tmail/mailet/LDAPMailingList.java
+++ b/tmail-backend/mailets/src/main/java/com/linagora/tmail/mailet/LDAPMailingList.java
@@ -34,18 +34,18 @@ import org.slf4j.LoggerFactory;
 import com.github.fge.lambdas.Throwing;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicate;
-import com.google.common.base.Splitter;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.unboundid.ldap.sdk.Attribute;
+import com.unboundid.ldap.sdk.DN;
 import com.unboundid.ldap.sdk.Filter;
 import com.unboundid.ldap.sdk.LDAPConnectionPool;
 import com.unboundid.ldap.sdk.LDAPException;
 import com.unboundid.ldap.sdk.LDAPSearchException;
+import com.unboundid.ldap.sdk.RDNNameValuePair;
 import com.unboundid.ldap.sdk.ResultCode;
 import com.unboundid.ldap.sdk.SearchResult;
 import com.unboundid.ldap.sdk.SearchResultEntry;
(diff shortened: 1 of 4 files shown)


tmail-backend/mailets/src/main/java/com/linagora/tmail/mailet/LDAPMailingList.java
```diff
@@ -449,18 +457,41 @@ SenderValidationPolicy domainRestrictedValidationPolicy() {
...
+
+    private String santinizeBusinessCategory(String ldapValue, String extractAttribute) {
+        if (ldapValue != null && !ldapValue.isEmpty()) {
+            // Split the businessCategory string by comma, map to key-value pairs, and collect to a MultiMap<String, String>
+            ImmutableListMultimap<String, String> businessAttributeMap = Splitter.on(',')
+                .trimResults()
+                .splitToStream(ldapValue)
+                .map(part -> part.trim().split("=", 2))
+                .filter(keyValue -> keyValue.length == 2) // Ensure valid key-ldapValue pairs
+                .collect(ImmutableListMultimap.toImmutableListMultimap(keyValue -> keyValue[0].trim(), keyValue -> keyValue[1].trim()));
```
chibenwa, line 489: Use unboundid DN class (or something similar) for the parsing?

tmail-backend/mailets/src/main/java/com/linagora/tmail/mailet/LDAPMailingList.java
```diff
@@ -65,6 +67,10 @@
  *  <li>userMailCacheDuration: Time during which one should keep entries into the user DN => mailAddress cache.</li>
  *  <li>mailAttributeForGroups: Attribute holding the mail address of a group. For easy testing this can be set to description
  *  but for production use a special LDAP schema needs to be crafted for using the mail attribute.</li>
+ *  <li>businessCategoryExtractAttribute: Optional string. Default to none. If configured, businessCategory would be
```
chibenwa, line 70: Not fan of the extra attribute to be honnest, I'd prefer we always move on with sanitizing.

This would simplify the configuration.

quantranhong1999: OK.

But I guess we can not do the same on James side.

chibenwa: IMO we can.

quantranhong1999: > Not fan of the extra attribute to be honnest, I'd prefer we always move on with sanitizing.

Do you want both normal and sanitize mode? Or only sanitize mode?

BTW on James side we can not just hardcode to sanitize the `cn`, no?

chibenwa: > BTW on James side we can not just hardcode to sanitize the cn, no?

matcher configuration is hard TBH as everything has to fit in one string.

We could have a system variable to disable that.

chibenwa: > Do you want both normal and sanitize mode? Or only sanitize mode?

Both.

I have a piece of code adressing my comments. Just running tests and I will push it.

tmail-backend/mailets/src/main/java/com/linagora/tmail/mailet/LDAPMailingList.java
quantranhong1999:
```java
    private String santinizeBusinessCategory(String ldapValue) throws LDAPException {
        if (ldapValue != null && !ldapValue.isEmpty()) {
            DN dn = new DN(ldapValue);

            // Iterate through the RDNs (Relative Distinguished Names)
            for (RDN rdn : dn.getRDNs()) {
                // Iterate over the attribute names within the RDN
                String[] attributeNames = rdn.getAttributeNames();
                String[] attributeValues = rdn.getAttributeValues();

                for (int i = 0; i < attributeNames.length; i++) {
                    if ("cn".equalsIgnoreCase(attributeNames[i])) {
                        return attributeValues[i];
                    }
                }
            }
        }
        return "";
    }
```

Works

chibenwa (approved):

PR #1215, opened 2024-10-03
[ISSUE-1163] Inject OpenPaaS user contacts into Tmail's auto-complete database

### LINKED ISSUE #1163: Tmail: Inject user contacts into auto-complete database
# Why

GIVEN I create abcdef@domain.tld in OpenPaaS
WHEN I compose an email in Twake mail and type abc
THEN it proposes abcdef@domain.tld as a potential recipient

# How

Listen to exchange `contacts:contact:add`

Given 

```
{
    "bookId": "65ae6175751dbd001b5799e8",
    "bookName": "contacts",
    "contactId": "fd9b3c98-fc77-4187-92ac-d9f58d400968",
    "userId": "65ae6175751dbd001b5799e8",
    "vcard": [
        "vcard",
        [
            [
                "version",
                {},
                "text",
                "4.0"
            ],
            [
                "uid",
                {},
                "text",
                "fd9b3c98-fc77-4187-92ac-d9f58d400968"
            ],
            [
                "fn",
                {},
                "text",
                "trst wqbvwerbwerb"
            ],
            [
                "n",
                {},
                "text",
                [
                    "wqbvwerbwerb",
                    "trst",
                    "",
                    "",
                    ""
                ]
            ],
            [
                "email",
                {
                    "type": "Work"
                },
                "text",
                "abcdef@domain.tld"
            ]
        ],
        []
    ]
}
```

Do a lookup on OpenPaaS to get the associated email 

```
# API to adapt but it looks like: (with the id instead of the mail)
curl -u "admin@open-paas.org:secret" https://backend.poc-mail-avocat.fr/api/users?email=c3po@poc-mail-avocat.fr
```

and index the mail address onto the relevant user search database.

Updates needs to be handled too:

```
{
    "bookId": "65ae6175751dbd001b5799e8",
    "bookName": "contacts",
    "contactId": "fd9b3c98-fc77-4187-92ac-d9f58d400968",
    "userId": "65ae6175751dbd001b5799e8",
    "vcard": [
        "vcard",
        [
            [
                "version",
                {},
                "text",
                "4.0"
            ],
            [
                "uid",
                {},
                "text",
                "fd9b3c98-fc77-4187-92ac-d9f58d400968"
            ],
            [
                "fn",
                {},
                "text",
                "trst wqbvwerbwerb"
            ],
            [
                "n",
                {},
                "text",
                [
                    "wqbvwerbwerb",
                    "trst",
                    "",
                    "",
                    ""
                ]
            ],
            [
                "email",
                {
                    "type": "Work"
                },
                "text",
                "abcdefg@domain.tld"
            ]
        ],
        []
    ]
}
```

# Definition of done

 - Names shall be used to set the mailbox display name of the recipient
 - When I send an email to someone it should not override the display name set by the calendar
 - Display name from the calendar should override the one of automatic indexing upon sending email (that likely to not have a display name)
 - Separate optional module, should share AMQP base configuration than https://github.com/linagora/tmail-backend/issues/1162

HoussemNasri:
Fix #1163


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class OpenPaasContactsConsumer {
+    private final Logger LOGGER = LoggerFactory.getLogger(OpenPaasContactsConsumer.class);
+
+    private static final String QUEUE_NAME = "contacts:contact:add";
+    private final ReactorRabbitMQChannelPool channelPool;
+    private Disposable consumeContactsDisposable;
+
+    // TODO: Create a separate RabbitMQ module for OpenPaaS communication so the injected channel pool
+    //  would be custom configured
```
chibenwa, line 24: I support this! A Lot!

We could...

Have a file named `openpaas.properties` which allow overriding RABBITMQ connection settings, if absent we can fallback to the default ChannelPool. It would be annotated with @Named("openpaas")

Then we can use it for OpenPaaS related communications in particular have a `OpenPaaSAMQPForwardAttribute` that extends `AmqpForwardAttribute and reuses this.

Actually such a task might be a good prelimiary worrk to this PR!

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+            Boolean.TRUE.equals(channelPool.getChannelMono()
+                .map(channel -> {
+                    try {
+                        channel.queueDeclarePassive(QUEUE_NAME);
+                        return true;
+                    }
+                    catch (IOException e) {
+                        return false;
+                    }
+                }).block());
```
chibenwa, line 50: That's a VERY good finding, I wasn't aware of this methods.

So if the queue already exist but with slightly different arguments does it throws? Can you please quickly test that? If no, then we have a solution to our major RabbitMQ migration nightmares Cc @QuangHoang1210 @Arsnael @vttranlina and we could use it more widely in the source code...

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+            LOGGER.warn("The queue '{}' is not present on the configured RabbitMQ server.", QUEUE_NAME);
+            return;
+        }
+
+        consumeContactsDisposable = channelPool.createReceiver()
+            .consumeAutoAck(QUEUE_NAME)
+            .map(delivery -> new String(delivery.getBody()))
+            .map(this::toContactUserAddedEvent)
+            .doOnError(
+                e -> LOGGER.error("Error while attempting to consume Contacts from OpenPaaS", e))
```
chibenwa, line 64:
```suggestion
            .doOnError(e -> LOGGER.error("Error while attempting to consume Contacts from OpenPaaS", e))
```

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+        // TODO: Maybe create queue when not exist? But we would have to make sure the queue config is in sync
+        //       between the publisher and consumer .e.g. if it's durable there it should be durable here.
+        if (!doesQueueExist) {
+            LOGGER.warn("The queue '{}' is not present on the configured RabbitMQ server.", QUEUE_NAME);
+            return;
+        }
+
+        consumeContactsDisposable = channelPool.createReceiver()
+            .consumeAutoAck(QUEUE_NAME)
+            .map(delivery -> new String(delivery.getBody()))
```
chibenwa, line 61: What is the charset?

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.io.IOException;
+import reactor.core.Disposable;
+import scala.NotImplementedError;
+
+
+public class OpenPaasContactsConsumer {
+    private final Logger LOGGER = LoggerFactory.getLogger(OpenPaasContactsConsumer.class);
+
+    private static final String QUEUE_NAME = "contacts:contact:add";
```
chibenwa, line 19: Please let's create our own queue for TMail and not attempt to reuse the OpenPaaS one. Otherwize we are stilling messages from OpenPaaS (and it might do valuable things with it)

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+
+        throw new NotImplementedError();
+    }
+
+    /**
+     * Start consuming.
+     */
+    public void start() {
+        boolean doesQueueExist =
+            Boolean.TRUE.equals(channelPool.getChannelMono()
```
chibenwa, line 41:
```suggestion
            channelPool.getChannelMono()
```

What is this heresy? IDE suggestions?

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+    private TmailContactUserAddedEvent toContactUserAddedEvent(String addedContactJSONMessage) {
+        // TODO: Extract relevant contact information using JsonPath
+
+        throw new NotImplementedError();
+    }
+
+    /**
+     * Start consuming.
+     */
+    public void start() {
```
chibenwa, line 39:
```suggestion
    public void startConsuming() {
```

Add `Clean code` on your must read list, and let's avoid needless comments.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.apache.james.backends.rabbitmq.ReactorRabbitMQChannelPool;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linagora.tmail.james.jmap.contact.TmailContactUserAddedEvent;
+
+import java.io.IOException;
+import reactor.core.Disposable;
+import scala.NotImplementedError;
```
chibenwa, line 13: I bet this do not pass the checkstyle

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,80 @@
...
+    public void start() {
+        boolean doesQueueExist =
+            Boolean.TRUE.equals(channelPool.getChannelMono()
+                .map(channel -> {
+                    try {
+                        channel.queueDeclarePassive(QUEUE_NAME);
+                        return true;
+                    }
+                    catch (IOException e) {
+                        return false;
```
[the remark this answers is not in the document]
chibenwa: method extraction to give a name to this non obvious operation

quantranhong1999 (commented):
Read it

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Test
+    void contactShouldBeIndexedWhenContactUserAddedMessage() {
+        sendMessage("""
+            {
+                "firstname" : "Alice",
+                "lastname"  : "Doe",
+                "email"     : "bob@domain.tld",
+                "username"  : "alice12"
+            }
+            """);
```
chibenwa, line 72: Not the format specified in the issue https://github.com/linagora/tmail-backend/issues/1163

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,89 @@
...
+        IntStream.range(0, 10).forEach(i -> sendMessage("BAD_PAYLOAD" + i));
+        TimeUnit.MILLISECONDS.sleep(100);
+
+        sendMessage("INVALID_PAYLOAD");
+
+        await().timeout(TEN_SECONDS).untilAsserted(() ->
+            assertThat(Flux.from(searchEngine.autoComplete(AccountId.fromString("bob@domain.tld"), "alice", 10))
+                .collectList().block())
+                .hasSize(1));
+    }
```
chibenwa, line 61: How  awaiting to have bob proves anything about not crashing the queue?

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumerTest.java
HoussemNasri: Yes, I wanted to get something working quickly, the format specified is hard to parse using Java, I plan to do it in Scala, but that would take time as I haven't done Scala in a while.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+        this.sender = sender;
+        this.commonRabbitMQConfiguration = commonRabbitMQConfiguration;
+        this.contactSearchEngine = contactSearchEngine;
+    }
+
+    @Override
+    public void start() {
+        Flux.concat(
+                sender.declareExchange(ExchangeSpecification.exchange(TOPIC)
+                    .durable(DURABLE)),
```
chibenwa, line 66: declare the exchange as fanout (because that's the settings openpaas uses.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+                sender.declareQueue(QueueSpecification
+                    .queue(QUEUE_NAME)
+                    .durable(evaluateDurable(DURABLE, commonRabbitMQConfiguration.isQuorumQueuesUsed()))),
```
chibenwa, line 69: Queue MUST always be durable.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
HoussemNasri: Where can I find the RabbitMQ config OpenPaaS uses?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+    private Disposable consumeContactsDisposable;
+
+    private final ReceiverProvider receiverProvider;
+    private final Sender sender;
+    private final RabbitMQConfiguration commonRabbitMQConfiguration;
+    private final EmailAddressContactSearchEngine contactSearchEngine;
+    private final Gson gson = new Gson();
+
+    // TODO: Create a separate RabbitMQ module for OpenPaaS communication so the injected channel pool
+    //  would be custom configured
```
chibenwa, line 50: Let's remove the todo and create a ticket instead?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,120 @@
...
+public class OpenPaasContactsConsumer implements Startable, Closeable {
+    public static final String TOPIC = "contacts:contact:add";
+    public static final String QUEUE_NAME = "ConsumeOpenPaasContactsQueue";
+    private Disposable consumeContactsDisposable;
+
+    private final ReceiverProvider receiverProvider;
+    private final Sender sender;
+    private final RabbitMQConfiguration commonRabbitMQConfiguration;
+    private final EmailAddressContactSearchEngine contactSearchEngine;
+    private final Gson gson = new Gson();
```
chibenwa, line 47: Use Jackson please !

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactMessage.java
```diff
@@ -0,0 +1,52 @@
+package com.linagora.tmail.james.jmap.contact;
+
+import java.util.Objects;
+
+public class OpenPaasContactMessage {
```
vttranlina, line 5: `record`will help us less boilerplate

tmail-backend/jmap/extensions/pom.xml
```diff
@@ -24,6 +24,16 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>team-mailboxes</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
```
chibenwa, line 28: Rather than creating all the code in `tmail-backend/jmap` wouldn't it be better to create a module named `tmail-backend/tmail-third-party/openpaas` ?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/jCardObjectDeserializer.java
```diff
@@ -0,0 +1,57 @@
...
+    public jCardObject deserialize(JsonParser p, DeserializationContext ctxt)
+        throws IOException, JacksonException {
+        JsonNode node = p.getCodec().readTree(p);
+
+        JsonNode jCardPropertiesArray = node.get(PROPERTIES_ARRAY_INDEX);
+        Iterator<JsonNode> propertiesIterator = jCardPropertiesArray.iterator();
+
+        String fn = "";
+        String email = "";
+        while (propertiesIterator.hasNext()) {
```
chibenwa, line 38: Let's work on streams 

(Iterators.toStream(...) )

And let's adopt a 'functional' style!

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/jCardObjectDeserializer.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Iterator;
+
+import com.fasterxml.jackson.core.JacksonException;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.google.common.base.Strings;
+
+public class jCardObjectDeserializer extends StdDeserializer<jCardObject> {
```
chibenwa, line 13: Nice !

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/jCardObjectDeserializer.java
```diff
@@ -0,0 +1,57 @@
...
+            } else if (EMAIL.equals(propertyName)) {
+                email = propertyNode.get(TEXT_PROPERTY_VALUE_INDEX).asText();
+            }
+        }
+
+        if (Strings.isNullOrEmpty(fn)) {
+            throw new RuntimeException("The fn property is required for a valid jCard object.");
+        } else if (Strings.isNullOrEmpty(email)) {
+            throw new RuntimeException("The email property is required for a valid jCard object.");
+        }
```
chibenwa, line 53: So first we can very well add a contact that do not have a `fn` in the contact DB so let's make the vcard field optional.

And second, let's reserve errors, especially runtimes to things that (hopefully) should not happen. Can I create a contact without email? IMO yes. In such case Twake mail should simply skip it.

I'd be making both fields of the jCard object optional and handle absebce in caller code.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/jCardObject.java
```diff
@@ -0,0 +1,29 @@
+package com.linagora.tmail.james.jmap.contact;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@JsonDeserialize(using = jCardObjectDeserializer.class)
+public record jCardObject(String fn, String email) {
```
chibenwa, line 6:
```suggestion
public record jCardObject(String fn, MailAddress email) {
```

Let's use strong types!

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -92,29 +96,43 @@ public Flux<AcknowledgableDelivery> delivery() {
 
     private void messageConsume(AcknowledgableDelivery ackDelivery, String messagePayload) {
         Mono.just(messagePayload)
-            .map(message -> gson.fromJson(message, OpenPaasContactMessage.class))
+            .<ContactAddedRabbitMqMessage>handle((message, sink) -> {
+                try {
+                    sink.next(objectMapper.readValue(message, ContactAddedRabbitMqMessage.class));
+                } catch (JsonProcessingException e) {
+                    sink.error(new RuntimeException(e));
```
chibenwa, line 103: Do a `nack` here?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -92,29 +96,43 @@ public Flux<AcknowledgableDelivery> delivery() {
...
+            })
             .handle((msg, sink) -> {
                 handleMessage(msg, sink);
                 ackDelivery.ack();
-            });
+            })
+            .onErrorResume(e -> {
+                LOGGER.error("Failed to consume OpenPaaS added contact message", e);
+                return Mono.empty();
+            }).subscribe();
```
chibenwa, line 113: Do not subscribe but make the method return a `Mono<Void>` (your operation) rather than `void`

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -92,29 +96,43 @@ public Flux<AcknowledgableDelivery> delivery() {
 
     private void messageConsume(AcknowledgableDelivery ackDelivery, String messagePayload) {
         Mono.just(messagePayload)
-            .map(message -> gson.fromJson(message, OpenPaasContactMessage.class))
+            .<ContactAddedRabbitMqMessage>handle((message, sink) -> {
```
chibenwa: We likely do not need that "just" and can inline the reactor transformations.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -92,29 +96,43 @@ public Flux<AcknowledgableDelivery> delivery() {
...
-            MailAddress mailAddress = new MailAddress(message.getEmail());
-            Mono.from(contactSearchEngine.index(AccountId.fromUsername(username),
-                new ContactFields(mailAddress, firstname, lastname))).block();
+            jCardObject jCardObject = contactAddedMessage.vcard();
+            String fullname = jCardObject.fn();
+            MailAddress mailAddress = new MailAddress(jCardObject.email());
+            Username emailAsUsername = Username.fromMailAddress(mailAddress);
+
+            Mono.from(contactSearchEngine.index(AccountId.fromUsername(emailAsUsername),
+                new ContactFields(mailAddress, fullname, ""))).block();
```
chibenwa, line 125: Do not block here but return the mono and compose it (`.flatMap` ?) in caller code.

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @AfterEach
+    void afterEach() throws IOException {
+        consumer.close();
+    }
+
+    @Test
+    void consumeMessageShouldNotCrashOnInvalidMessages() {
+        IntStream.range(0, 10).forEach(i -> sendMessage("BAD_PAYLOAD" + i));
```
chibenwa, line 51: Then make sure you can consume a regular message and get the expected outcome.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private void handleMessage(ContactAddedRabbitMqMessage contactAddedMessage, SynchronousSink<Object> sink) {
+        LOGGER.info("Consumed jCard object message: {}", contactAddedMessage);
+        try {
+            jCardObject jCardObject = contactAddedMessage.vcard();
+            String fullname = jCardObject.fn();
+            MailAddress mailAddress = new MailAddress(jCardObject.email());
+            Username emailAsUsername = Username.fromMailAddress(mailAddress);
+
+            Mono.from(contactSearchEngine.index(AccountId.fromUsername(emailAsUsername),
```
chibenwa, line 124: Bob creates a contact named "alice" and we add here alice contact into alice address book.

What would be needed here is to resolve the email address of bob (to compute his AccountId) and then add alice contact in it.

For this we need the call to openpaas user API

We can hide such an operation behind an interface to isolate this from the rest of the code here.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
HoussemNasri: Ticket created: https://github.com/linagora/tmail-backend/issues/1223

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumerTest.java
HoussemNasri: Fixed.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactMessage.java
HoussemNasri: Done.

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumerTest.java
HoussemNasri: Done.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasWebClient.java
```diff
@@ -0,0 +1,41 @@
...
+        return client.get()
+            .uri(String.format("/users?email=%s", email))
+            .response()
+            .cast(OpenPaasUserResponse[].class)
+            .map(userResponseArray -> Arrays.stream(userResponseArray).toList());
+    }
+
+    private String basicAuthenticationHeaderValue() {
+        String userPassword = openPaasConfiguration.getWebClientUser() + ":" + openPaasConfiguration.getWebClientPassword();
+        byte[] base64UserPassword = Base64.getEncoder().encode(userPassword.getBytes(StandardCharsets.UTF_8));
```
chibenwa, line 36: We can store this as a field to avoid to recompute it on every request

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
chibenwa: image

Because I can check it in real world RabbitMQ admin ;-)

chibenwa (commented):
Cool but IMO we would deserve to have dedicated tests for `OpenPaaSWebClient` class.

(Test in isolation for this)

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasWebClient.java
```diff
@@ -0,0 +1,50 @@
...
+import java.time.Duration;
+import java.util.Base64;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import reactor.core.publisher.Mono;
+import reactor.netty.http.client.HttpClient;
+
+public class OpenPaasWebClient {
```
chibenwa, line 13:
```suggestion
public class OpenPaasRestClient {
```

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/jCardObjectDeserializer.java
HoussemNasri: > Can I create a contact without email? IMO yes

True, fn is required according to the specification, but without the email, there's nothing we can do with it.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,164 @@
...
+            }).subscribe();
+    }
+
+    private Mono<Void> handleMessage(ContactAddedRabbitMqMessage contactAddedMessage) {
+        LOGGER.info("Consumed jCard object message: {}", contactAddedMessage);
+
+        String openPaasOwnerId = contactAddedMessage.userId();
+        return openPaasRestClient.getUserById(openPaasOwnerId)
+            .map(OpenPaasUserResponse::preferredEmail)
+            .mapNotNull(ownerEmail -> {
```
chibenwa, line 125: mapNotNull ? When can we have null? If we can produce null data producer code needs to be modified.

tmail-backend/jmap/extensions/pom.xml
```diff
@@ -24,6 +24,16 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>team-mailboxes</artifactId>
         </dependency>
+        <dependency>
```
chibenwa, line 27: Why are there modifs in jmap extensions pom file? Accident?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/api/OpenPaasUserResponse.java
```diff
@@ -0,0 +1,16 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record OpenPaasUserResponse(@JsonProperty("id") String id,
+                                   @JsonProperty("firstname") String firstname,
+                                   @JsonProperty("lastname") String lastname,
+                                   @JsonProperty("preferredEmail") String preferredEmail,
+                                   @JsonProperty("emails") List<String> emails,
+                                   @JsonProperty("main_phone") String mainPhone,
+                                   @JsonProperty("displayName") String displayName) {
```
chibenwa, line 15: Strip uneeded fields?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/JCardObjectDeserializer.java
```diff
@@ -0,0 +1,72 @@
...
+        throws IOException {
+        JsonNode node = p.getCodec().readTree(p);
+
+        JsonNode jCardPropertiesArray = node.get(PROPERTIES_ARRAY_INDEX);
+        Map<String, String> jCardProperties =
+            collectJCardProperties(jCardPropertiesArray.iterator());
+
+        if (!jCardProperties.containsKey(FN)) {
+            throw new RuntimeException("The FN field is required according to specification.");
+        }
```
chibenwa, line 47: I believe we should be more lenient (be lenient on what you accept strict on what you send)

Can we make FN optional in our datmodel? Or position it to an empty string?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/JCardObjectDeserializer.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private static Map<String, String> collectJCardProperties(Iterator<JsonNode> propertiesIterator) {
+        return Iterators.toStream(propertiesIterator)
+            .map(JCardObjectDeserializer::getPropertyKeyValuePair)
+            .filter(pair -> pair != ImmutablePair.<String, String>nullPair())
+            .collect(Collectors.toMap(Pair::getKey, Pair::getValue));
+    }
+
+    private static ImmutablePair<String, String> getPropertyKeyValuePair(JsonNode propertyNode) {
```
chibenwa, line 59:
```suggestion
    private static ImmutablePair<String, String> getPropertyKeyValuePair(JsonNode propertyNode) {
```
```suggestion
    private static Optional<ImmutablePair<String, String>> getPropertyKeyValuePair(JsonNode propertyNode) {
```

chibenwa: Or Stream<...> so that we can then use flatMap

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,159 @@
...
+import reactor.rabbitmq.Receiver;
+import reactor.rabbitmq.Sender;
+
+
+public class OpenPaasContactsConsumer implements Startable, Closeable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(OpenPaasContactsConsumer.class);
+
+    public static final String EXCHANGE_NAME = "contacts:contact:add";
+    public static final String QUEUE_NAME = "ConsumeOpenPaasContactsQueue";
+    private Disposable consumeContactsDisposable;
```
chibenwa, line 49: Move it after final fields.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private final ReceiverProvider receiverProvider;
+    private final Sender sender;
+    private final RabbitMQConfiguration commonRabbitMQConfiguration;
+    private final EmailAddressContactSearchEngine contactSearchEngine;
+    private final ObjectMapper objectMapper = new ObjectMapper();
+    private final OpenPaasRestClient openPaasRestClient;
+
+    // TODO: Create a separate RabbitMQ module for OpenPaaS communication so the injected channel pool
+    //  would be custom configured
```
chibenwa, line 59:
```suggestion
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,159 @@
...
+    @Inject
+    public OpenPaasContactsConsumer(@Named(EmailAddressContactInjectKeys.AUTOCOMPLETE) ReceiverProvider receiverProvider,
+                                    @Named(EmailAddressContactInjectKeys.AUTOCOMPLETE) Sender sender,
```
chibenwa, line 62: Inject a RabbitMQCHannelPool

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,159 @@
...
+        this.openPaasRestClient = openPaasRestClient;
+    }
+
+    public void start() {
+        Flux.concat(
+                sender.declareExchange(ExchangeSpecification.exchange(EXCHANGE_NAME)
+                    .durable(DURABLE).type(BuiltinExchangeType.FANOUT.getType())),
+                sender.declareQueue(QueueSpecification
+                    .queue(QUEUE_NAME)
+                    .durable(DURABLE)),
```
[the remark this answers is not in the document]
chibenwa: We likly need also to create the deadletter queue

tmail-backend/tmail-third-party/openpaas/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+    </parent>
+
+    <artifactId>tmail-openpaas</artifactId>
+    <name>Twake Mail :: Third Party :: OpenPaaS</name>
+    <description>OpenPaaS integration for Twake Mail</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>jmap-extensions</artifactId>
```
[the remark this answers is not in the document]
chibenwa: Link of the ticket?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
chibenwa: > Inject a RabbitMQCHannelPool

Not done

tmail-backend/tmail-third-party/openpaas/pom.xml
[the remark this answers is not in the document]
HoussemNasri: I'm working on it, but I'm not sure about the reason behind this change. Will it improve compile times, for example?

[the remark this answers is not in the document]
HoussemNasri: https://github.com/linagora/tmail-backend/issues/1254

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,134 @@
...
+import reactor.core.publisher.Mono;
+import reactor.rabbitmq.OutboundMessage;
+
+import com.linagora.tmail.OpenPaasConfiguration;
+import com.linagora.tmail.api.OpenPaasRestClient;
+import com.linagora.tmail.api.OpenPaasServerExtension;
+import com.linagora.tmail.james.jmap.contact.EmailAddressContactSearchEngine;
+import com.linagora.tmail.james.jmap.contact.InMemoryEmailAddressContactSearchEngine;
+
+class OpenPaasContactsConsumerTest {
```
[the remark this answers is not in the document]
chibenwa: Please where are the edge case tested?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,159 @@
...
+                if (contactMailAddressOpt.isEmpty()) {
+                    return Mono.empty();
+                }
+
+                AccountId ownerAccountId =
+                    AccountId.fromUsername(Username.fromMailAddress(ownerMailAddress));
+
+                return Mono.from(contactSearchEngine.index(ownerAccountId,
+                        new ContactFields(contactMailAddressOpt.get(), contactFullname, "")
+                    )).block();
```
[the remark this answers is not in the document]
HoussemNasri: Agree, removed the block here but I have to keep it when creating the exchanges/queues at the beginning.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,184 @@
...
+            });
+
+        return combineOptionals(contactFullnameOpt, contactMailAddressOpt,
+            (contactFullname, contactMailAddress) ->
+                new ContactFields(contactMailAddress, contactFullname, contactFullname));
+    }
+
+    private static <T,K,V> Optional<V> combineOptionals(Optional<T> opt1, Optional<K> opt2, BiFunction<T, K, V> f) {
+        return opt1.flatMap(t1 -> opt2.map(t2 -> f.apply(t1, t2)));
+    }
```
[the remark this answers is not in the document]
HoussemNasri: We map two optionals into one and apply a `.map` on them. The result optional is present only if both input optionals are present. It is like .zip in scala.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,184 @@
...
+                sender.declareExchange(ExchangeSpecification.exchange(DEAD_LETTER_EXCHANGE)
+                    .durable(DURABLE)),
+                sender.declareQueue(QueueSpecification
+                    .queue(DEAD_LETTER_QUEUE)
+                    .durable(DURABLE)
+                    .arguments(commonRabbitMQConfiguration.workQueueArgumentsBuilder().build())),
+                sender.declareQueue(QueueSpecification
+                    .queue(DEAD_LETTER_QUEUE)
+                    .durable(DURABLE)
+                    .arguments(commonRabbitMQConfiguration.workQueueArgumentsBuilder().build())),
```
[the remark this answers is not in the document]
HoussemNasri: Rebasing gone wrong, fixed.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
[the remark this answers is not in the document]
HoussemNasri: Didn't get to that part, had a bug in one of the existing test cases, so I prioritized fixing that first.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        consumer.start();
+    }
+
+    @AfterEach
+    void afterEach() throws IOException {
+        consumer.close();
+    }
+
+    @Test
+    void consumeMessageShouldNotCrashOnInvalidMessages() throws InterruptedException {
```
HoussemNasri, line 66: Testcase is failing

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -120,9 +119,15 @@ private Mono<EmailAddressContact> messageConsume(AcknowledgableDelivery ackDeliv
...
             .flatMap(this::handleMessage)
-            .doOnSuccess(input -> ackDelivery.ack())
-            .doOnError(e -> ackDelivery.nack(REQUEUE_ON_NACK))
-            .onErrorResume(e -> Mono.error(new RuntimeException("Failed to consume OpenPaaS added contact message", e)));
+            .doOnSuccess(result -> {
+                LOGGER.warn("Consumed contact successfully '{}'", result);
+                ackDelivery.ack();
+            })
+            .onErrorResume(error -> {
+                LOGGER.error("Error when consume message '{}'", messagePayload, error);
```
chibenwa, line 127: Please let's not log the full payload...

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+        return Mono.just(messagePayload)
+            .map(this::parseContactAddedRabbitMqMessage)
+            .flatMap(this::handleMessage)
+            .doOnSuccess(result -> {
+                LOGGER.warn("Consumed contact successfully '{}'", result);
+                ackDelivery.ack();
+            })
+            .onErrorResume(error -> {
+                LOGGER.error("Error when consume message '{}'", messagePayload, error);
+                ackDelivery.nack(REQUEUE_ON_NACK);
```
chibenwa, line 128: REQUEUE_ON_NACK should be true and !REQUEUE_ON_NACK on here 

(second edition)

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+    private ContactAddedRabbitMqMessage parseContactAddedRabbitMqMessage(String message) {
+        try {
+            return objectMapper.readValue(message, ContactAddedRabbitMqMessage.class);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException("Failed to parse ContactAddedRabbitMqMessage", e);
+        }
+    }
+
+    private Mono<EmailAddressContact> handleMessage(ContactAddedRabbitMqMessage contactAddedMessage) {
+        LOGGER.info("Consumed jCard object message: {}", contactAddedMessage);
```
chibenwa, line 142: LOGGER.trace

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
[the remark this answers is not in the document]
chibenwa: 1. I know reading code, but this seems overly complex for the task at hand.
2. variables names are not acceptable
3. I am not even sure this is what we want... If the name is missing, is it legit to index the mail address? Likely yes...

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+                    return Optional.of(new MailAddress(contactEmail));
+                } catch (AddressException e) {
+                    LOGGER.warn("Invalid contact email address: {}", contactEmail, e);
+                    return Optional.empty();
+                }
+            });
+
+        return combineOptionals(contactFullnameOpt, contactMailAddressOpt,
+            (contactFullname, contactMailAddress) ->
+                new ContactFields(contactMailAddress, contactFullname, contactFullname));
```
chibenwa, line 169: Does it make sense to index twice the full name?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    private static <T,K,V> Optional<V> combineOptionals(Optional<T> opt1, Optional<K> opt2, BiFunction<T, K, V> f) {
+        return opt1.flatMap(t1 -> opt2.map(t2 -> f.apply(t1, t2)));
+    }
+
+    @Override
+    public void close() throws IOException {
+        if (consumeContactsDisposable != null) {
```
chibenwa, line 178: `Optional.ifPresent(...)` as we like functional constructs?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+        return openPaasRestClient.retrieveMailAddress(contactAddedMessage.userId())
+            .map(ownerMailAddress -> AccountId.fromUsername(Username.fromMailAddress(ownerMailAddress)))
+            .flatMap(ownerAccountId ->
+                Mono.justOrEmpty(contactFieldsOpt)
+                    .flatMap(contactFields -> doAddContact(ownerAccountId, contactFields)));
+    }
+
+    private Mono<EmailAddressContact> doAddContact(AccountId ownerAccountId, ContactFields contactFields) {
+        return Mono.from(contactSearchEngine.index(ownerAccountId, contactFields));
+    }
```
chibenwa, line 153: Not sure this is worth the pain of extracting a method.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private Mono<EmailAddressContact> handleMessage(ContactAddedRabbitMqMessage contactAddedMessage) {
+        LOGGER.info("Consumed jCard object message: {}", contactAddedMessage);
+        Optional<ContactFields> contactFieldsOpt = toContactFields(contactAddedMessage.vcard());
+        return openPaasRestClient.retrieveMailAddress(contactAddedMessage.userId())
+            .map(ownerMailAddress -> AccountId.fromUsername(Username.fromMailAddress(ownerMailAddress)))
+            .flatMap(ownerAccountId ->
+                Mono.justOrEmpty(contactFieldsOpt)
+                    .flatMap(contactFields -> doAddContact(ownerAccountId, contactFields)));
```
chibenwa, line 148: While the type aligns and it likely works, it is not strictly equivalent on the execution side: flatMap is an expensive construct that brings asynchrony into play: it is generally better to keep reactive pipeline simpler, and handle logic resolution upfront if possible (code fragment I proposed and that is now lost in the comment history) but if we want we could save one flatMap with:

```suggestion
            .flatMap(ownerAccountId ->
                contactFieldsOpt.map(contactFields -> doAddContact(ownerAccountId, contactFields))
                   .orElse(Mono.empty());
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+        try {
+            return objectMapper.readValue(message, ContactAddedRabbitMqMessage.class);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException("Failed to parse ContactAddedRabbitMqMessage", e);
+        }
+    }
+
+    private Mono<EmailAddressContact> handleMessage(ContactAddedRabbitMqMessage contactAddedMessage) {
+        LOGGER.info("Consumed jCard object message: {}", contactAddedMessage);
+        Optional<ContactFields> contactFieldsOpt = toContactFields(contactAddedMessage.vcard());
```
chibenwa, line 143:
```suggestion
        Optional<ContactFields> maybeContactFields = toContactFields(contactAddedMessage.vcard());
```

opt is not in my english vocabulary, sorry

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+        return openPaasRestClient.retrieveMailAddress(contactAddedMessage.userId())
```
chibenwa, line 144: We are still retrieving the owner mail address for later discovering that we do not need it as the contact do not have a usable mail address...

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+            .flatMap(ownerAccountId ->
+                Mono.justOrEmpty(contactFieldsOpt)
+                    .flatMap(contactFields -> doAddContact(ownerAccountId, contactFields)));
+    }
+
+    private Mono<EmailAddressContact> doAddContact(AccountId ownerAccountId, ContactFields contactFields) {
+        return Mono.from(contactSearchEngine.index(ownerAccountId, contactFields));
+    }
+
+    private Optional<ContactFields> toContactFields(JCardObject jCardObject) {
```
chibenwa, line 155: We could extract this (and unit test it?) as part of the JCardObject class?

(jCardObject.asContactFields()) ... a

And thus simplify the current class?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+                ackDelivery.ack();
+            })
+            .onErrorResume(error -> {
+                LOGGER.error("Error when consume message '{}'", messagePayload, error);
+                ackDelivery.nack(REQUEUE_ON_NACK);
+                return Mono.empty();
+            });
+    }
+
+    private ContactAddedRabbitMqMessage parseContactAddedRabbitMqMessage(String message) {
```
chibenwa, line 133: Why the `RabbitMQMessage` suffix? Is it usefull?

While we are at it I would carry over the `byte[]` onto there and have the object mapper diorectly reading the value from bytes and not from string.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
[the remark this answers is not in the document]
chibenwa: Please review the changeset after pushing.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,182 @@
...
+    private final ReceiverProvider receiverProvider;
+    private final Sender sender;
+    private final RabbitMQConfiguration commonRabbitMQConfiguration;
+    private final EmailAddressContactSearchEngine contactSearchEngine;
+    private final ObjectMapper objectMapper = new ObjectMapper();
+    private final OpenPaasRestClient openPaasRestClient;
+
+    @Inject
+    public OpenPaasContactsConsumer(@Named(EmailAddressContactInjectKeys.AUTOCOMPLETE) ReceiverProvider receiverProvider,
+                                    @Named(EmailAddressContactInjectKeys.AUTOCOMPLETE) Sender sender,
```
chibenwa, line 64:
```suggestion
    public OpenPaasContactsConsumer(RabbitMQChannelPool channelPool, 
```

Still not done.

BTW we will figure out the injections later.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
chibenwa: Maybe we can paste here the error so that reviewers do not need to run the test themselves in order to figure out what is going on?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
HoussemNasri: Okay 😅 will do it, but for the record, I'm not a fan of this negative booleans-are-bad idea.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
HoussemNasri: No, it does not, I changed it to index the fullname as the firstname and leave the surname empty. We can't just split by space the full name and put one part as a first name and the other as a surname as some fullnames could contain a prefix like 'Dr.' which would make the fullname containing more than one space.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
HoussemNasri: It's fixed now.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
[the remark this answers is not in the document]
HoussemNasri: Agreed, I changed it to index even when the fullname is missing but the mail address is present.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
HoussemNasri: How about on success?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
HoussemNasri: There isn't any isDurable() method in commonRabbitMQConfiguration, should I add one?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/contact/OpenPaasContactsConsumer.java
HoussemNasri: Not sure how...

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
HoussemNasri: Done.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
HoussemNasri: I think it is useful for code organization. For example, to search for all RabbitMQ messages used across the app I could just search for classes that have that suffix.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,184 @@
...
+                    .arguments(commonRabbitMQConfiguration.workQueueArgumentsBuilder().build())),
+                sender.declareQueue(QueueSpecification
+                    .queue(DEAD_LETTER_QUEUE)
+                    .durable(DURABLE)
+                    .arguments(commonRabbitMQConfiguration.workQueueArgumentsBuilder().build())),
+                sender.declareQueue(QueueSpecification
+                    .queue(QUEUE_NAME)
+                    .durable(DURABLE)
+                    .arguments(commonRabbitMQConfiguration.workQueueArgumentsBuilder()
+                    .put("x-dead-letter-exchange", DEAD_LETTER_EXCHANGE)
```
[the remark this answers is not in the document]
HoussemNasri: Hmmm, an exchange and a queue are two different concepts, as far as I understand. How would this work?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/JCardObject.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.james.core.MailAddress;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.google.common.base.Preconditions;
+import com.linagora.tmail.james.jmap.contact.ContactFields;
+
+@JsonDeserialize(using = JCardObjectDeserializer.class)
+public record JCardObject(Optional<String> fnOpt, Optional<String> emailOpt) {
```
HoussemNasri, line 16: Should we use MailAddress here or it's fine as is? And throw an exception at the deserializer.

chibenwa: Of course using stronger types in the data model indeed is always better. If we think this can easily be achieved then we shall go for it.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/ContactAddedRabbitMqMessage.java
```diff
@@ -0,0 +1,21 @@
...
+
+public record ContactAddedRabbitMqMessage(String bookId, String bookName, String contactId,
+                                          String userId, JCardObject vcard) {
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    static ContactAddedRabbitMqMessage fromJSON(byte[] jsonBytes) {
+        String message = new String(jsonBytes, StandardCharsets.UTF_8);
+        try {
+            return objectMapper.readValue(message, ContactAddedRabbitMqMessage.class);
```
chibenwa, line 16:
```suggestion
        try {
            return objectMapper.readValue(jsonBytes, ContactAddedRabbitMqMessage.class);
```

?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,151 @@
...
+            receiver -> receiver.consumeManualAck(QUEUE_NAME),
+            Receiver::close);
+    }
+
+    private Mono<EmailAddressContact> messageConsume(AcknowledgableDelivery ackDelivery, byte[] messagePayload) {
+        return Mono.just(messagePayload)
+            .map(ContactAddedRabbitMqMessage::fromJSON)
+            .flatMap(this::handleMessage)
+            .doOnSuccess(result -> {
+                LOGGER.info("Consumed contact successfully '{}'", result);
```
chibenwa, line 108:
```suggestion
                LOGGER.debug("Consumed contact successfully '{}'", result);
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,151 @@
...
+            })
+            .onErrorResume(error -> {
+                LOGGER.error("Error when consume message", error);
+                ackDelivery.nack(!REQUEUE_ON_NACK);
+                return Mono.empty();
+            });
+    }
+
+    private Mono<EmailAddressContact> handleMessage(ContactAddedRabbitMqMessage contactAddedMessage) {
+        LOGGER.debug("Consumed jCard object message: {}", contactAddedMessage);
```
chibenwa, line 119:
```suggestion
        LOGGER.trace("Consumed jCard object message: {}", contactAddedMessage);
```

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,378 @@
...
+    private OpenPaasContactsConsumer consumer;
+
+    @BeforeEach
+    void setup() throws URISyntaxException {
+        OpenPaasRestClient restClient = new OpenPaasRestClient(
+            new OpenPaasConfiguration(
+                openPaasServerExtension.getBaseUrl(),
+                OpenPaasServerExtension.GOOD_USER(),
+                OpenPaasServerExtension.GOOD_PASSWORD())
+        );
```
chibenwa, line 57:
```suggestion
                OpenPaasServerExtension.GOOD_PASSWORD()));
```

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,378 @@
...
+    }
+
+    /*
+    * The automatic contact indexing is triggered when you send or receive a message
+    * from someone, then their contact info be automatically indexed in the contacts' search engine.
+    * */
+    @Test
+    void automaticContactIndexingShouldNotOverrideContactInfoFromOpenPaas() {
+        //
+    }
```
chibenwa, line 353: ?

chibenwa (commented):
We are getting closer ;-)

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,151 @@
...
+    public static final String QUEUE_NAME = "openpaas-contacts-queue";
+    public static final String DEAD_LETTER = QUEUE_NAME + "-dead-letter";
+
+    private Disposable consumeContactsDisposable;
+    private final ReceiverProvider receiverProvider;
+    private final Sender sender;
+    private final RabbitMQConfiguration commonRabbitMQConfiguration;
+    private final EmailAddressContactSearchEngine contactSearchEngine;
+    private final ObjectMapper objectMapper = new ObjectMapper();
+    private final OpenPaasRestClient openPaasRestClient;
```
chibenwa, line 53:
```suggestion
    private final ReceiverProvider receiverProvider;
    private final Sender sender;
    private final RabbitMQConfiguration commonRabbitMQConfiguration;
    private final EmailAddressContactSearchEngine contactSearchEngine;
    private final ObjectMapper objectMapper = new ObjectMapper();
    private final OpenPaasRestClient openPaasRestClient;
    private Disposable consumeContactsDisposable;
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -0,0 +1,151 @@
...
+        return maybeContactFields.map(
+                openPaasContact -> openPaasRestClient.retrieveMailAddress(contactAddedMessage.userId())
+                    .map(this::getAccountIdFromMailAddress)
+                    .flatMap(ownerAccountId -> indexContactIfNeeded(ownerAccountId, openPaasContact)))
+            .orElse(Mono.empty());
+    }
+
+    private Mono<EmailAddressContact> indexContactIfNeeded(AccountId ownerAccountId, ContactFields openPaasContact) {
+        return Mono.from(contactSearchEngine.get(ownerAccountId, openPaasContact.address()))
+            .onErrorResume(e -> Mono.empty())
```
chibenwa, line 130:
```suggestion
            .onErrorResume(ContactNotFoundException.class, e -> Mono.empty())
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/ContactAddedRabbitMqMessage.java
HoussemNasri: How about the encoding?

HoussemNasri: Looked into the source code, they use an automatic encoding detection algorithm. It seems to be performant as it only checks some number of prefix bits to make the prediction.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/contact/openpaas/OpenPaasContactsConsumer.java
HoussemNasri: https://github.com/linagora/tmail-backend/issues/1223

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
HoussemNasri: Now that we have defined the expected behavior. i.e. OpenPaas is the single source of truth for contacts, this test case become irrelevant.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,134 @@
...
+                  ]
+               ]
+            }
+            """);
+
+        await().timeout(TEN_SECONDS).untilAsserted(() ->
+            assertThat(
+                Flux.from(searchEngine.autoComplete(AccountId.fromString(OpenPaasServerExtension.ALICE_EMAIL()), "jhon", 10))
+                    .collectList().block())
+                .hasSize(1));
```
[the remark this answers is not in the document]
HoussemNasri: Should this be applied to all test cases? The test case that verifies the successful addition of a contact already includes a more specific assertion.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/contact/OpenPaasContactsConsumerTest.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    /*
+    * The automatic contact indexing is triggered when you send or receive a message
+    * from someone, then their contact info be automatically indexed in the contacts' search engine.
+    * */
+    @Test
+    void automaticContactIndexingShouldNotOverrideContactInfoFromOpenPaas() {
+        //
+    }
```
chibenwa, line 352:
```suggestion
```

chibenwa (approved):
Should be more or less good to me

quantranhong1999 (approved):
LGTM

PR #1229, opened 2024-10-11
[IMPROVEMENT] Set default `-Dextra.props=/root/conf/jvm.properties` for TMail apps

quantranhong1999:
- Better convenient as this is our common path to use too.
- Match default behavior with James apps

### DIFF
--- a/tmail-backend/apps/distributed/src/main/conf/jvm.properties
+++ b/tmail-backend/apps/distributed/src/main/conf/jvm.properties
@@ -5,7 +5,7 @@
 # my.property=whatever
 
 # Required to locate Cassandra driver configuration
-config.file=conf/cassandra-driver.conf
+config.file=/root/conf/cassandra-driver.conf
 
 # (Optional). String (size, integer + size units, example: `12 KIB`, supported units are bytes KIB MIB GIB TIB). Defaults to 100KIB.
 # This governs the threshold MimeMessageInputStreamSource relies on for storing MimeMessage content on disk.
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,7 @@
                                 <jvmFlag>-XX:+ExitOnOutOfMemoryError</jvmFlag>
                                 <jvmFlag>-Dlogback.configurationFile=/root/conf/logback.xml</jvmFlag>
                                 <jvmFlag>-Dworking.directory=/root/</jvmFlag>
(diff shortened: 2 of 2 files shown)


quantranhong1999:
I make it as a draft to validate if it is a breaking change or not.

tmail-backend/apps/distributed/src/main/conf/jvm.properties
```diff
@@ -5,7 +5,7 @@
 # my.property=whatever
 
 # Required to locate Cassandra driver configuration
-config.file=conf/cassandra-driver.conf
+config.file=/root/conf/cassandra-driver.conf
```
quantranhong1999, line 8: I have checked and on tmail.linagora.com, TWP env and CNB we use `config.file=/root/conf/cassandra-driver.conf` too.

The thing I do not understand: on MU we set path for `jvm.properties`:
```yaml
    environment:
      - JAVA_TOOL_OPTIONS= ... -Dextra.props=/root/conf/jvm.properties
```

Then we use relative path to load `cassandra-driver.conf` in `jvm.properties`:
```
config.file=conf/cassandra-driver.conf
```
But it seems still work on MU... That is what I don't understand.

chibenwa (approved):

tmail-backend/apps/distributed/src/main/conf/jvm.properties
Arsnael: Maybe because the working directory is set to `/root/`? => <jvmFlag>-Dworking.directory=/root/</jvmFlag>

Relative path `conf/cassandra-driver.conf` from working directory `/root/` is the same as absolute path `/root/conf/cassandra-driver.conf` ? 

Would be my guess :)

Arsnael (approved):

tmail-backend/apps/distributed/src/main/conf/jvm.properties
quantranhong1999: > Relative path conf/cassandra-driver.conf from working directory /root/ is the same as absolute path /root/conf/cassandra-driver.conf

Theoretically it should work but it was not on my local. And I see a similar fix for that too (fixed by you :p):
`JAMES-3668 Add absolute paths for extra properties file in docker image`

vttranlina (approved):

PR #1235, opened 2024-10-14
ISSUE-1231 SecondaryBlobStoreDAO should read failover when the primary S3 is down

quantranhong1999:


chibenwa (approved):

Arsnael (approved):

hungphan227:
sr Quan, I forgot to tell you: Regarding the bucket errot, we should return Mono.empty in S3BlobStoreDAO when NoSuchBucketException occur

quantranhong1999:
> sr Quan, I forgot to tell you: Regarding the bucket error, we should return Mono.empty in S3BlobStoreDAO when NoSuchBucketException occur

Thank FYI, but it is not related to this ticket, is it?

Issue #1232, opened 2024-10-14
[S3 reliability] Use FileBackedOutputStream to back SecondaryBlobStore::save


quantranhong1999:
## Why

```
Indeed we are forced to copy the inputstream as the two daos can't read the inputStream.

However  IOUtils.toByteArray(inputStream) operation is blocking: please to it onto a Scheduler.boundedElastic()...

Moreover this API might be called with objects that might not fit in memory. I would welcome the use of FileBackedOutputStream that uses a file if the input is > 100MB.
```

cf https://github.com/linagora/tmail-backend/pull/1214/files#r1791802833

## DoD

Finish the TODO:
https://github.com/linagora/tmail-backend/blob/e8dbaea8299d34042d3dbadaac924e00a1fba8fa/tmail-backend/blob/secondary-blob-store/src/main/java/com/linagora/tmail/blob/secondaryblobstore/SecondaryBlobStoreDAO.java#L89

quantranhong1999:
https://github.com/linagora/tmail-backend/pull/1256

PR #1256, opened 2024-10-22
ISSUE-1232 SecondaryBlobStoreDAO should use FileBackedOutputStream to back save()

quantranhong1999:
resolve https://github.com/linagora/tmail-backend/issues/1232


tmail-backend/blob/secondary-blob-store/src/main/java/com/linagora/tmail/blob/secondaryblobstore/SecondaryBlobStoreDAO.java
```diff
@@ -48,6 +78,7 @@ public boolean isSuccess() {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SecondaryBlobStoreDAO.class);
     private static final Set<RegistrationKey> NO_REGISTRATION_KEYS = ImmutableSet.of();
+    private static final int FILE_THRESHOLD = 1024 * 100;
```
chibenwa, line 81: Let's go for 

```suggestion
    private static final int FILE_THRESHOLD = 100 * 1024 * 1024;
```

Arsnael (approved):

chibenwa (approved):

Issue #1223, opened 2024-10-08
Create a separate Guice module for OpenPaaS communication


HoussemNasri:
# Description

We should have a separate Guice module for OpenPaaS communication to be able to inject custom RabbitMQ configuration that is specific to OpenPaas.

chibenwa:
> Depends on: https://github.com/linagora/tmail-backend/pull/1215

Not really. IMO the too tasks are independant.

We can also write a `OpenPaaSAmqpForwardAttribute` mailet that is exactly `AmqpForwardAttribute` but configured using OpenPaaS configuration.

chibenwa:
We decided with Houssem to not implement fallback for distributed-app to the main RabbitMQChannelPool.

We can live with 2 pools (we already live with 4 of them!)

PR #1261, opened 2024-10-24
Create a separate Guice module for OpenPaaS communication

HoussemNasri:
Closes https://github.com/linagora/tmail-backend/issues/1223

## TODO
- [x] Test `OpenPaasConfiguration` (see `RabbitMQConfigurationTest`)
- [x] Test `AmqpURI`
- [x] ~~<Make the "add contacts" exchange configurable.~~ https://github.com/linagora/tmail-backend/issues/1280
- [x] Sort pom dependecies
- [x] Throw exception when RabbitMQ URI not configured in openpaas.proeprties.


tmail-backend/apps/distributed/src/main/java/com/linagora/tmail/james/app/DistributedServer.java
```diff
@@ -312,7 +313,8 @@ protected void configure() {
             new TeamMailboxModule(),
             new TMailMailboxSortOrderProviderModule(),
             new TmailEventModule(),
-            new TmailEventDeadLettersModule());
+            new TmailEventDeadLettersModule(),
+            new OpenPaasModule());
```
quantranhong1999, line 317:
```
22:23:34.258 [ERROR] c.l.t.OpenPaasModule - Could not find configuration file 'openpaas.properties'
22:23:34.271 [ERROR] o.a.j.GuiceJamesServer - Fatal error while starting James
java.io.FileNotFoundException: openpaas
```

Likely we should make the OpenPaaS module optional?

chibenwa: Fully agree.

If unspecified we shall fallback to the main rabbitMQ connection and rely on / vhost.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasConfiguration.java
```diff
@@ -1,6 +1,78 @@
...
+import java.net.URI;
+import java.util.Optional;
+
+import org.apache.commons.configuration2.Configuration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Strings;
+
+public record OpenPaasConfiguration(Optional<AmqpUri> maybeRabbitMqUri, Optional<URI> openpaasApiUri, Optional<String> maybeAdminUser, Optional<String> maybeAdminPassword) {
```
chibenwa, line 12:
```suggestion
public record OpenPaasConfiguration(Optional<AmqpUri> maybeRabbitMqUri, URI openpaasApiUri, String maybeAdminUser, String maybeAdminPassword) {
```

Forgive me but this extension cannot work without OpenPaaS infos?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasConfiguration.java
```diff
@@ -1,6 +1,78 @@
...
+            LOGGER.debug("OpenPaas API URI not specified.");
+            return Optional.empty();
+        }
+
+        try {
+            return Optional.of(URI.create(openPaasApiUri));
+        } catch (IllegalArgumentException e) {
+            LOGGER.error("Invalid OpenPaas API URI in openpaas.properties.");
+            return Optional.empty();
+        }
```
chibenwa, line 55: Propagate the exception so that the admin fixes the conf glitch please

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasConfiguration.java
```diff
@@ -1,6 +1,78 @@
...
+        if (Strings.isNullOrEmpty(rabbitMqUri)) {
+            LOGGER.debug("RabbitMQ URI not defined in openpaas.properties, falling back to rabbitmq.properties.");
+            return Optional.empty();
+        }
+
+        try {
+            return AmqpUri.from(URI.create(rabbitMqUri)).asOptional();
+        } catch (IllegalArgumentException e) {
+            LOGGER.error("Invalid RabbitMQ URI in openpaas.properties, falling back to rabbitmq.properties.");
+            return Optional.empty();
```
chibenwa, line 39: Propagate the exception so that the admin fixes the conf glitch please

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasModule.java
```diff
@@ -0,0 +1,105 @@
...
+    @Provides
+    @Named(OPENPAAS_CONFIGURATION_NAME)
+    @Singleton
+    public Configuration providePropertiesConfiguration(PropertiesProvider propertiesProvider) throws ConfigurationException {
+        try {
+            return propertiesProvider.getConfiguration(OPENPAAS_CONFIGURATION_NAME);
+        } catch (FileNotFoundException e) {
+            LOGGER.error("Could not find configuration file '{}.properties'",
+                OPENPAAS_CONFIGURATION_NAME);
+            throw new RuntimeException(e);
```
chibenwa, line 50: We shall likely do a module chooser to start or not the openpaas extension?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasModule.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Provides
+    public OpenPaasRestClient provideOpenPaasRestCLient(OpenPaasConfiguration openPaasConfiguration) {
+        return new OpenPaasRestClient(openPaasConfiguration);
+    }
+
+    @Provides
+    @Named(OPENPAAS_INJECTION_KEY)
+    @Singleton
+    public RabbitMQConfiguration provideRabbitMQConfiguration(OpenPaasConfiguration openPaasConfiguration, RabbitMQConfiguration fallbackRabbitMQConfiguration) {
```
chibenwa, line 68: Here we would connect twice to RabbitMQ.

Is there a way to connect only once when we fallback?

Also on top of memory-app there would be no fallback as we do not use RabbitMQ at all...

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasModule.java
HoussemNasri: Could you confirm if this is accurate: The extension loads only if `openpaas.properties` is present in the classpath. If any OpenPaas configurations (e.g., apiURL or admin details) are missing, it will fail. If the RabbitMQ URI is not configured, it falls back to `rabbitmq.properties` (**question**: for the fallback, should we use all configurations in rabbitmq.properties or just the URI?)

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasModule.java
HoussemNasri: > Here we would connect twice to RabbitMQ.

Can you please elaborate on this part?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasConfiguration.java
HoussemNasri: Not it cannot, done.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasModule.java
chibenwa: > If any OpenPaas configurations (e.g., apiURL or admin details) are missing, it will fail.

Wrong.

For `Distributed` if AMQP settings are missing we are falling back to the mail RabbitMQChannelPool

For `Memory` we are failing

(This likely means we need a little `DistributedOpenPaaSAMQPFallbackModule` and `MemoryOpenPaaSAMQPFallbackModule` to encode this difference)

For the fallback we use the RabbitMQChannelPool object already instanciated, let's not read rabbitmq.properties ourselves.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/AmqpUri.java
```diff
@@ -86,4 +86,22 @@ private RabbitMQConfiguration.ManagementCredentials parseUserInfo(String userInf
                 .join(passwordParts)
                 .toCharArray());
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
```
chibenwa, line 94: We can likely get rid of this micro optimization

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/AmqpUri.java
```diff
@@ -86,4 +86,22 @@ private RabbitMQConfiguration.ManagementCredentials parseUserInfo(String userInf
                 .join(passwordParts)
                 .toCharArray());
     }
+
+    @Override
+    public boolean equals(Object o) {
```
chibenwa, line 91: Shall be `final`. 

(Is this tested by equals-verifier?)

tmail-backend/apps/distributed/src/main/java/com/linagora/tmail/james/app/DistributedOpenPaasAmqpFallbackModule.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.james.backends.rabbitmq.RabbitMQConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+import com.linagora.tmail.AmqpUri;
+import com.linagora.tmail.configuration.OpenPaasConfiguration;
+
+public class DistributedOpenPaasAmqpFallbackModule extends AbstractModule {
```
chibenwa, line 17: We need tests for both the regular behaviour (when rabbitmq uri is configured) and the fallback, for both distributed and mamory App.

In order to do that:
 - we will not create the `openpaas.properties` file
 - edit the configuration of the JamesServerExtension we are using in order to turn on the OpenPaaS extension
 - we will then need a (server) overide where we define the OpenPaaS extension configuration. It ;s ok as we already have solid tests for configuration parsing.
 
 So 4 little tests to write!

tmail-backend/apps/distributed/src/main/java/com/linagora/tmail/james/app/DistributedOpenPaasAmqpFallbackModule.java
```diff
@@ -0,0 +1,33 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(
+        DistributedOpenPaasAmqpFallbackModule.class);
+
+    @Provides
+    @Named(OPENPAAS_INJECTION_KEY)
+    @Singleton
+    public RabbitMQConfiguration provideRabbitMQConfiguration(
```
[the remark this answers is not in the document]
HoussemNasri: I don't think we can. The provider takes another provider annotated with @Named("openpaas"), if we annotate this provider with the same annotation, which is the attended use case we will get a BINDING_ALREADY_SET error. We could skip the annotation but that will still conflict with the default RabbitMQChannelPool provider declared elsewhere.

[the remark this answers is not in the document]
chibenwa: > I don't think we can.

Create more intermediary abstractions to break the dependency loops. Or use other named annotations.

Please find something. Be inventive and try things.

tmail-backend/apps/memory/src/test/java/com/linagora/tmail/james/app/MemoryServerWithOpenPaasTest.java
```diff
@@ -0,0 +1,75 @@
...
+            .overrideWith(new LinagoraTestJMAPServerModule())
+            .overrideWith(binder -> Multibinder.newSetBinder(binder, GuiceProbe.class).addBinding().to(MailboxManagerClassProbe.class))
+            .overrideWith(binder -> binder.bind(OpenPaasConfiguration.class).toInstance(
+                new OpenPaasConfiguration(
+                    Optional.empty(),
+                    URI.create("https://127.0.0.1"),
+                    "admin",
+                    "admin"
+                )
+            )))
```
chibenwa, line 46:
```
                    "admin",
                    "admin"))))
```

Those categories of comments are not worth my time. Please pay attention!

tmail-backend/apps/memory/src/test/java/com/linagora/tmail/james/app/MemoryServerWithOpenPaasTest.java
```diff
@@ -0,0 +1,75 @@
...
+        .build();
+
+    @BeforeAll
+    static void setUp() {
+        System.setSecurityManager(new NoExitSecurityManager());
+    }
```
chibenwa, line 52: WHY?

tmail-backend/apps/memory/src/test/java/com/linagora/tmail/james/app/MemoryServerWithoutOpenPaasTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.james.utils.GuiceProbe;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import com.google.inject.multibindings.Multibinder;
+import com.linagora.tmail.encrypted.MailboxConfiguration;
+import com.linagora.tmail.encrypted.MailboxManagerClassProbe;
+import com.linagora.tmail.module.LinagoraTestJMAPServerModule;
+
+public class MemoryServerWithoutOpenPaasTest {
```
chibenwa, line 16: I think all the other tests check what happens without the openpaas module enabled.

As it stands this test is useless.

We shall rather check the exception we get when we try to start tmail with the openpaas extension without configuring RabbitMQ...

tmail-backend/apps/distributed/src/test/java/com/linagora/tmail/james/app/DistributedServerWithoutOpenPaasTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.inject.multibindings.Multibinder;
+import com.linagora.tmail.OpenPaasModuleChooserConfiguration;
+import com.linagora.tmail.blob.guice.BlobStoreConfiguration;
+import com.linagora.tmail.combined.identity.UsersRepositoryClassProbe;
+import com.linagora.tmail.encrypted.MailboxConfiguration;
+import com.linagora.tmail.encrypted.MailboxManagerClassProbe;
+import com.linagora.tmail.module.LinagoraTestJMAPServerModule;
+
+public class DistributedServerWithoutOpenPaasTest {
```
chibenwa, line 18: Idem this test is useless.

chibenwa:
To test that RabbitMQ and OpenPaaS rabbitMQ are different we could leverage a "Probe"

We can create a TestProbe with injections on `RabbitMQChannelPool` and `@Named("openpaas") RabbitMQChannelPool` that exposes a boolean method to show if those are the sames or not.

tmail-backend/apps/memory/src/test/java/com/linagora/tmail/james/app/MemoryServerWithOpenPaasTest.java
HoussemNasri: On failure, the server does not throw an exception it exits by calling `System.exit`. Here we are trying to test that the memory server fails when the OpenPaas module is enabled and there is no rabbitmq channel pool to fallback to.

chibenwa: Here is is worth a comment?

chibenwa, line 52: @AfterAll remember to unset the security manager or test isolation is broken.

tmail-backend/apps/memory/src/test/java/com/linagora/tmail/james/app/MemoryServerWithOpenPaasTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    void serverShouldThrowWithOpenPaasEnabled() throws Exception {
+        try {
+            jamesServerExtension.start(null);
+            fail("James server didn't exit or fail!");
+        } catch (Exception e) {
+            LOGGER.error("Expected failure", e);
+        }
+    }
```
chibenwa, line 62: disable auto start on the extention, get a non-started guice-james-server as a parameter and start it manually in the tests

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/AmqpUri.java
HoussemNasri: No, not yet. I'm currently testing AmqpURI but haven't pushed the changes yet because I'm encountering some issues with the URI class in the Java standard. For instance, BAD_URI is considered a valid URI and does not throw an exception.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/OpenPaasRabbitMQExtension.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.james.GuiceModuleTestExtension;
+import org.apache.james.backends.rabbitmq.DockerRabbitMQ;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+
+import com.google.inject.Module;
+
+public class OpenPaasRabbitMQExtension implements GuiceModuleTestExtension {
+    private final DockerRabbitMQ dockerRabbitMQ = DockerRabbitMQ.withoutCookie();
```
chibenwa, line 12: IMO this extension is not needed we could have 2 connections to the main RabbitMQ extension and we would still be able to assert if we want that the mail RabbitMQChannelPool is distinct from the one annotated with `@Named("openpaas")`

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/AmqpUriTest.java
```diff
@@ -0,0 +1,112 @@
...
+class AmqpUriTest {
+    private static Stream<Arguments> goodAmqpURIs() {
+        return Stream.of(
+            Arguments.of("amqp://guest:guest@localhost:5672/"),
+            Arguments.of("amqp://user:password@host:port/vhost"),
+            Arguments.of("amqps://user:password@securehost:5671/securevhost"), // Using AMQPS for secure connection
+            Arguments.of("amqp://@localhost:5672/"),
+            Arguments.of("amqp://:password@host"),
+            Arguments.of("amqp://user@host")
+            );
```
chibenwa, line 29:
```suggestion
            Arguments.of("amqp://user@host"));
```

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/AmqpUriTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private static Stream<Arguments> badAmqpURIs() {
+        return Stream.of(
+            Arguments.of("http://guest:guest@localhost:5672/"), // Wrong protocol
+            Arguments.of("amqp://user:pass@host:5672/extra/path"),      // Extra path element
+            Arguments.of("BAD_URI")                                     // Just bad
+        );
```
chibenwa, line 37:
```suggestion
            Arguments.of("BAD_URI"));                                     // Just bad
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/AmqpUri.java
```diff
@@ -1,107 +1,157 @@
...
             return false;
         }
         return Objects.equals(uri, amqpUri.uri) &&
-               Objects.equals(managementCredentials, amqpUri.managementCredentials) &&
+               Objects.equals(userInfo, amqpUri.userInfo) &&
                Objects.equals(vhost, amqpUri.vhost);
     }
 
     @Override
     public int hashCode() {
```
chibenwa, line 145:
```suggestion
    public final int hashCode() {
```

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/AmqpUriTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("badAmqpURIs")
+    void testBadAmpqUriString(String amqpUri) {
+        Assertions.assertThatThrownBy(() -> AmqpUri.from(amqpUri));
+    }
+
+    @Test
+    void shouldRespectBeanContract() throws URISyntaxException, NoSuchAlgorithmException, KeyManagementException {
```
chibenwa, line 54:
```suggestion
    void shouldRespectBeanContract() throws Exception {
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/AmqpUri.java
```diff
@@ -1,107 +1,157 @@
...
-        return Objects.hash(uri, managementCredentials, vhost);
+        return Objects.hash(uri, userInfo, vhost);
+    }
+
+    @Override
+    public String toString() {
+        return "AmqpUri{" +
```
chibenwa, line 151: MoreObjects.toStringHelpers()...

tmail-backend/apps/distributed/src/main/java/com/linagora/tmail/james/app/DistributedOpenPaasAmqpFallbackModule.java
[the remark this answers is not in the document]
HoussemNasri: It was decided not to implement the fallback in this PR, keeping it open so I can return to it when working on the fallback.

tmail-backend/apps/distributed/pom.xml
```diff
@@ -93,6 +93,11 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>tmail-mailets</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>tmail-openpaas</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
chibenwa, line 99: Move the version declaration into `tmail-backend/pom.xml`, use `${project.version}` there and here just declare the dependency without the version.

tmail-backend/apps/distributed/src/test/java/com/linagora/tmail/james/app/DistributedServerWithOpenPaasRabbitMqConfiguredTest.java
```diff
@@ -0,0 +1,46 @@
...
+            .workingDirectory(tmpDir)
+            .configurationFromClasspath()
+            .searchConfiguration(SearchConfiguration.openSearch())
+            .usersRepository(UsersRepositoryModuleChooser.Implementation.COMBINED)
+            .eventBusKeysChoice(EventBusKeysChoice.REDIS)
+            .openPassModuleChooserConfiguration(OpenPaasModuleChooserConfiguration.ENABLED)
+            .build())
+        .server(configuration -> DistributedServer.createServer(configuration)
+            .overrideWith(new LinagoraTestJMAPServerModule())
+            .overrideWith(binder -> Multibinder.newSetBinder(binder, GuiceProbe.class).addBinding().to(UsersRepositoryClassProbe.class)))
```
chibenwa, line 33: Is this module override needed?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/OpenPaasModule.java
```diff
@@ -0,0 +1,100 @@
...
+    @Provides
+    @Named(OPENPAAS_CONFIGURATION_NAME)
+    @Singleton
+    public Configuration providePropertiesConfiguration(PropertiesProvider propertiesProvider) throws ConfigurationException {
+        try {
+            return propertiesProvider.getConfiguration(OPENPAAS_CONFIGURATION_NAME);
+        } catch (FileNotFoundException e) {
+            LOGGER.error("Could not find configuration file '{}.properties'",
+                OPENPAAS_CONFIGURATION_NAME);
+            throw new RuntimeException(e);
```
chibenwa, line 50: Please just return disabled here: this is optional.

HoussemNasri: What is optional? we are loading the OpenPaasModule using a module chooser, this code is only executed if we explicitly enable the module, now that the fallback is not implemented, we fail when the openpaas.properties file is not found.

tmail-backend/apps/distributed/src/test/java/com/linagora/tmail/james/app/DistributedServerWithOpenPaasRabbitMqConfiguredTest.java
HoussemNasri: Probably not, we don't use jmap in this test. The test passes even after removing the module. It is hard to tell for sure which modules or extensions are needed and which are not.

chibenwa (approved):

Arsnael (approved):

PR #1304, opened 2024-11-13
small version updates and fixups

florentos17:
when trying to run demos and run james, i encountered several small issues that needed the fixups i suggest here.

### DIFF
--- a/dockerfiles/jenkins-runner/Dockerfile
+++ b/dockerfiles/jenkins-runner/Dockerfile
@@ -17,7 +17,7 @@ RUN set -x && \
   rm -rf /var/lib/apt/lists/*
 
 # Install Maven 3.6.3
-ENV MAVEN_VERSION 3.6.3
+ENV MAVEN_VERSION 3.9.9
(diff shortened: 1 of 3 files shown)


demo/tmail/mailetcontainer.xml
```diff
@@ -96,13 +96,6 @@
             <mailet match="All" class="com.linagora.tmail.mailets.TmailLocalDelivery">
                 <consume>false</consume>
             </mailet>
-
-            <!-- Put the OpenAIMailet after LocalDelivery so the GPT reply would come after the asking question -->
-            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.AIBotMailet">
-                <apiKey>demo</apiKey>
-                <botAddress>gpt@tmail.com</botAddress>
-                <model>gpt-4o-mini</model>
-            </mailet>
```
florentos17, line 105: this mailet caused the server to fail at booting, removing that fixed the problem for me. but i don't know if we should remove it entirely though.

chibenwa (approved):

Arsnael (approved):

PR #1316, opened 2024-11-18
ISSUE-1307 Add secondary blob store suffix onto its default namespace

chibenwa:
Failure to do so causes it to be double applied.


quantranhong1999 (approved):
Thanks!

Arsnael (approved):

Issue #1275, opened 2024-10-30
Synchronise contact delete


chibenwa:
### What

When I delete a contact from my address book

Then it must no longer show up in my auto-complete suggestions

### How 

Listen to `contacts:contact:delete`

Sample message:

```
{
    "bookId": "5f50b136bdaffe0026290b3d",
    "bookName": "e01850bd-d013-47e4-8b1a-02e462f0d4df",
    "contactId": "7cc163d4-bc25-4479-a969-e1662529f4ad",
    "userId": "5f50b136bdaffe0026290b3d",
    "vcard": [
        "vcard",
        [
            [
                "version",
                {},
                "text",
                "4.0"
            ],
            [
                "uid",
                {},
                "text",
                "7cc163d4-bc25-4479-a969-e1662529f4ad"
            ],
            [
                "fn",
                {},
                "text",
                "(Ms.) Vo Cao Ky Duyen"
            ],
            [
                "n",
                {},
                "text",
                [
                    "Duyen",
                    "(Ms.) Vo Cao Ky",
                    "",
                    "",
                    ""
                ]
            ],
            [
                "org",
                {},
                "text",
                "AEON Vietnam Academy"
            ],
            [
                "role",
                {},
                "text",
                "Training Executive"
            ],
            [
                "email",
                {
                    "type": "Work"
                },
                "text",
                "mailto:duyen.vo@aeon.com.vn"
            ],
            [
                "tel",
                {
                    "type": "Work"
                },
                "uri",
                "(028) 6288 7711 – Ext : 649"
            ]
        ],
        []
    ]
}
```

Issue #1274, opened 2024-10-30
Synchronise contact update


chibenwa:
## What

```
GIVEN a contact Aalice -> aalice@toto.com
WHEN I update the contact mail address to Alice -> alice@toto.com
THEN I search *ali*
AND the autocomplete shall return *alice*
```

## How

Listen to `contacts:contact:update` queue

Sample message:

```
{
    "bookId": "5f50a663bdaffe002629099c",
    "bookName": "contacts",
    "contactId": "bae79596-f6fc-4812-b9e0-4eafe32c2f3d",
    "userId": "5f50a663bdaffe002629099c",
    "vcard": [
        "vcard",
        [
            [
                "version",
                {},
                "text",
                "4.0"
            ],
            [
                "uid",
                {},
                "text",
                "bae79596-f6fc-4812-b9e0-4eafe32c2f3d"
            ],
            [
                "fn",
                {},
                "text",
                "dd"
            ],
            [
                "n",
                {},
                "text",
                [
                    "",
                    "dd",
                    "",
                    "",
                    ""
                ]
            ],
            [
                "categories",
                {},
                "text",
                "vwrbwrbr"
            ],
            [
                "email",
                {
                    "type": "Work"
                },
                "text",
                "mailto:d@d.com"
            ]
        ],
        []
    ]
}
```

Update the autocomplete database accordingly.

PR #1367, opened 2024-12-04
synchronize tmail's auto-complete database after openpaas user update/delete operations

florentos17:
tackling the Synchronise contact delete #1275 and Synchronise contact update #1274 issues

i basically created two more exchanges and two more queues, and depending on the queue, the message is treated differently (new `updateContact` and `deleteContact` similar to the existing `indexContactIfNeeded`).

since the key for these `index`, `update`, `delete` operations is the `accountId` which is obtained through the mail address, the entry is correctly updated when the address does not change (previous entry replaced by the new one, cf the test `contactShouldBeUpdatedWhenContactUserUpdatedMessageWithSameAddress`), whereas if the address is updated then both the old and the new entries are there (cf the test `newContactShouldBeAddedWhenContactUserUpdatedMessageWithUpdatedAddress`)

### DIFF
--- a/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/ContactRabbitMqMessage.java
+++ b/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/ContactRabbitMqMessage.java
@@ -5,7 +5,7 @@ import java.io.IOException;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
-public record ContactAddedRabbitMqMessage(String bookId, String bookName, String contactId,
+public record ContactRabbitMqMessage(String bookId, String bookName, String contactId,
                                           String userId, JCardObject vcard) {
 
     private static final ObjectMapper objectMapper = new ObjectMapper();
@@ -14,11 +14,11 @@ public record ContactAddedRabbitMqMessage(String bookId, String bookName, String
         objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
     }
 
-    static ContactAddedRabbitMqMessage fromJSON(byte[] jsonBytes) {
+    static ContactRabbitMqMessage fromJSON(byte[] jsonBytes) {
         try {
-            return objectMapper.readValue(jsonBytes, ContactAddedRabbitMqMessage.class);
+            return objectMapper.readValue(jsonBytes, ContactRabbitMqMessage.class);
         } catch (IOException e) {
-            throw new RuntimeException("Failed to parse ContactAddedRabbitMqMessage", e);
+            throw new RuntimeException("Failed to parse ContactRabbitMqMessage", e);
         }
     }
 }
--- a/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
+++ b/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
@@ -40,9 +40,15 @@ import reactor.rabbitmq.Sender;
 public class OpenPaasContactsConsumer implements Startable, Closeable {
     private static final Logger LOGGER = LoggerFactory.getLogger(OpenPaasContactsConsumer.class);
     private static final boolean REQUEUE_ON_NACK = true;
-    public static final String EXCHANGE_NAME = "contacts:contact:add";
-    public static final String QUEUE_NAME = "openpaas-contacts-queue";
-    public static final String DEAD_LETTER = QUEUE_NAME + "-dead-letter";
+    public static final String EXCHANGE_NAME_ADD = "contacts:contact:add";
+    public static final String EXCHANGE_NAME_UPDATE = "contacts:contact:update";
+    public static final String EXCHANGE_NAME_DELETE = "contacts:contact:delete";
+    public static final String QUEUE_NAME_ADD = "openpaas-contacts-queue-add";
+    public static final String QUEUE_NAME_UPDATE = "openpaas-contacts-queue-update";
+    public static final String QUEUE_NAME_DELETE = "openpaas-contacts-queue-delete";
+    public static final String DEAD_LETTER_ADD = "openpaas-contacts-queue-add-dead-letter";
+    public static final String DEAD_LETTER_UPDATE = "openpaas-contacts-queue-update-dead-letter";
+    public static final String DEAD_LETTER_DELETE = "openpaas-contacts-queue-delete-dead-letter";
 
     private final ReceiverProvider receiverProvider;
     private final Sender sender;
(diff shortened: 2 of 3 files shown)


tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/ContactRabbitMqMessage.java
```diff
@@ -0,0 +1,19 @@
+package com.linagora.tmail.contact;
+
+import java.io.IOException;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public record ContactRabbitMqMessage(String bookId, String bookName, String contactId,
```
chibenwa, line 7: Remember me to verify on friday that indeed message format is always the same...

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -42,9 +42,13 @@
...
-    public static final String EXCHANGE_NAME = "contacts:contact:add";
-    public static final String QUEUE_NAME = "openpaas-contacts-queue";
-    public static final String DEAD_LETTER = QUEUE_NAME + "-dead-letter";
+    public static final String EXCHANGE_NAME_ADD = "contacts:contact:add";
+    public static final String EXCHANGE_NAME_UPDATE = "contacts:contact:update";
+    public static final String EXCHANGE_NAME_DELETE = "contacts:contact:delete";
+    public static final String QUEUE_NAME_ADD = "openpaas-contacts-queue-add";
+    public static final String QUEUE_NAME_UPDATE = "openpaas-contacts-queue-update";
+    public static final String QUEUE_NAME_DELETE = "openpaas-contacts-queue-delete";
+    public static final String DEAD_LETTER = "openpaas-contacts-queue-dead-letter";
```
chibenwa, line 51: With 3 queue but only one dead letter how are we going to sort things out easily from deadletter and reprocess from correct location?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -117,13 +127,18 @@ private Mono<EmailAddressContact> messageConsume(AcknowledgableDelivery ackDeliv
...
-                openPaasContact -> openPaasRestClient.retrieveMailAddress(contactAddedMessage.userId())
+                openPaasContact -> openPaasRestClient.retrieveMailAddress(contactMessage.userId())
                     .map(this::getAccountIdFromMailAddress)
-                    .flatMap(ownerAccountId -> indexContactIfNeeded(ownerAccountId, openPaasContact)))
+                    .flatMap(ownerAccountId -> switch (queue) {
+                        case QUEUE_NAME_ADD -> indexContactIfNeeded(ownerAccountId, openPaasContact);
+                        case QUEUE_NAME_UPDATE -> updateContact(ownerAccountId, openPaasContact);
+                        case QUEUE_NAME_DELETE -> deleteContact(ownerAccountId, openPaasContact);
+                        default -> Mono.empty();
+                    }))
```
chibenwa, line 141: That's uggly, i'd prefer we have indirection with a lambda: in the consume, we declare how to consume a message for each queue and pass it down the stack and get rid of that tricky tricky switch.

BTW default option should have been `throw new Exception("Oh my god UNSUPPORTED");`

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -141,6 +156,28 @@ private Mono<EmailAddressContact> indexContactIfNeeded(AccountId ownerAccountId,
...
+        return Mono.from(contactSearchEngine.get(ownerAccountId, openPaasContact.address()))
+            .onErrorResume(ContactNotFoundException.class, e -> {
+                LOGGER.warn("Contact not found for update: {}", openPaasContact.address());
+                return Mono.empty();
+            })
+            .switchIfEmpty(
+                Mono.from(contactSearchEngine.index(ownerAccountId, openPaasContact)))
+            .flatMap(existingContact ->
+                Mono.from(contactSearchEngine.update(ownerAccountId, openPaasContact)));
+    }
```
chibenwa, line 169:
```suggestion
    private Mono<EmailAddressContact> updateContact(AccountId ownerAccountId, ContactFields openPaasContact) {
        return Mono.from(contactSearchEngine.get(ownerAccountId, openPaasContact.address()))
            .flatMap(existingContact -> Mono.from(contactSearchEngine.update(ownerAccountId, openPaasContact)));
            .onErrorResume(ContactNotFoundException.class, e -> {
                LOGGER.warn("Contact not found for update: {}", openPaasContact.address());
                return Mono.from(contactSearchEngine.index(ownerAccountId, openPaasContact));
            })
    }
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -141,6 +156,28 @@ private Mono<EmailAddressContact> indexContactIfNeeded(AccountId ownerAccountId,
...
+
+    private Mono<Void> deleteContact(AccountId ownerAccountId, ContactFields openPaasContact) {
+        return Mono.from(contactSearchEngine.get(ownerAccountId, openPaasContact.address()))
+            .onErrorResume(ContactNotFoundException.class, e -> {
+                LOGGER.warn("Contact not found for deletion: {}", openPaasContact);
+                return Mono.empty();
+            })
+            .flatMap(existingContact ->
+                Mono.from(contactSearchEngine.delete(ownerAccountId, openPaasContact.address())));
+    }
```
chibenwa, line 179: Or just blind delete: why do I need the initial read for?

chibenwa:
@HoussemNasri  please next time review this before I do.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -67,45 +74,51 @@ public OpenPaasContactsConsumer(@Named(OPENPAAS_INJECTION_KEY) ReactorRabbitMQCh
...
+        startExchange(EXCHANGE_NAME_ADD, QUEUE_NAME_ADD, DEAD_LETTER_ADD);
+        startExchange(EXCHANGE_NAME_UPDATE, QUEUE_NAME_UPDATE, DEAD_LETTER_UPDATE);
+        startExchange(EXCHANGE_NAME_DELETE, QUEUE_NAME_DELETE, DEAD_LETTER_DELETE);
+    }
+
+    public void startExchange(String exchange, String queue, String deadLetter) {
         Flux.concat(
-                sender.declareExchange(ExchangeSpecification.exchange(EXCHANGE_NAME)
+                sender.declareExchange(ExchangeSpecification.exchange(exchange)
                     .durable(DURABLE).type(BuiltinExchangeType.FANOUT.getType())),
```
HoussemNasri, line 85: Are we certain FANOUT is the exchange for all exchanges?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -117,16 +130,32 @@ private Mono<EmailAddressContact> messageConsume(AcknowledgableDelivery ackDeliv
...
     }
 
-    private Mono<EmailAddressContact> handleMessage(ContactAddedRabbitMqMessage contactAddedMessage) {
-        LOGGER.trace("Consumed jCard object message: {}", contactAddedMessage);
-        Optional<ContactFields> maybeContactFields = contactAddedMessage.vcard().asContactFields();
+    private Mono<?> handleMessage(ContactRabbitMqMessage contactMessage, String queue) {
+        LOGGER.trace("Consumed jCard object message: {}", contactMessage);
+        Optional<ContactFields> maybeContactFields = contactMessage.vcard().asContactFields();
+        ContactHandler contactHandler = contactHandlers.get(queue);
+        if (contactHandlers.get(queue) == null) {
```
HoussemNasri, line 137: We should probably move this at the beginning of the method.

HoussemNasri: And change the exception message message to something more descriptive. LLMs can come handy in this case.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -117,16 +130,32 @@ private Mono<EmailAddressContact> messageConsume(AcknowledgableDelivery ackDeliv
...
+                    .flatMap(ownerAccountId -> contactHandler.handle(ownerAccountId, openPaasContact)))
             .orElse(Mono.empty());
     }
 
+    @FunctionalInterface
+    private interface ContactHandler {
+        Mono<?> handle(AccountId ownerAccountId, ContactFields openPaasContact);
+    }
+
+    private final Map<String, ContactHandler> contactHandlers = Map.of(
```
HoussemNasri, line 153: OpenPaasContactsConsumer is doing too much and have many reasons to change. We should probably move the logic inside these methods into another class, similar to `com.linagora.tmail.james.jmap.contact.EmailAddressContactMessageHandler`.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
chibenwa: Yes that's reasonnable

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
chibenwa: The case of being asked to hanlde a queue  that do not exist should not even be a thing.
We should by design always cary on the information of what to do without using side bands to pass the information.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
```diff
@@ -112,7 +115,7 @@ public Flux<AcknowledgableDelivery> delivery(String queue) {
             Receiver::close);
     }
 
-    private Mono<Object> messageConsume(AcknowledgableDelivery ackDelivery, byte[] messagePayload, String queue) {
+    private Mono<?> messageConsume(AcknowledgableDelivery ackDelivery, byte[] messagePayload, String queue) {
```
chibenwa, line 118:
```suggestion
    private Mono<?> messageConsume(AcknowledgableDelivery ackDelivery, byte[] messagePayload, Consumer<ContactRabbitMqMessage> treatment) {
```

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/contact/OpenPaasContactsConsumer.java
chibenwa: Unsure of the benefits cost of a split: I like this central place.

IMO let's keep things simple now...

florentos17:
this time i passed the "treatment" method from the start so no "unknown queue" issue
i simply used a functional interface `ContactHandler` instead of the suggested `Consumer` but it' does not change much

florentos17:
i squashed the commits and resolved the conflicts

chibenwa (approved):

Arsnael (approved):

PR #1377, opened 2024-12-05
Introduce StandaloneEventAttendanceRepository

HoussemNasri:
Closes https://github.com/linagora/tmail-backend/issues/1326

### DIFF
--- a/tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
+++ b/tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
@@ -1,0 +1,64 @@
+package com.linagora.tmail.james.jmap;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import jakarta.mail.Flags;
+
+import org.apache.james.mailbox.FlagsBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+
+public enum AttendanceStatus {
+    Accepted("$accepted"),
+    Declined("$rejected"),
+    Tentative("$tentativelyaccepted"),
+    NeedsAction("$needs-action");
+
+    private static final Set<String> EVENT_ATTENDANCE_FLAGS = Arrays.stream(values())
+        .map(AttendanceStatus::getUserFlag)
+        .collect(Collectors.toSet());
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AttendanceStatus.class);
+
+    private final String userFlag;
+
+    AttendanceStatus(String userFlag) {
+        this.userFlag = userFlag;
+    }
+
+    public String getUserFlag() {
+        return userFlag;
+    }
+
+    public static Optional<AttendanceStatus> fromUserFlag(String userFlag) {
+        Preconditions.checkNotNull(userFlag);
+        return Arrays.stream(AttendanceStatus.values())
+            .filter(status -> status.userFlag.equals(userFlag))
+            .findFirst();
+    }
+
+    public static Optional<AttendanceStatus> fromMessageFlags(Flags flags) {
+        List<AttendanceStatus> eventAttendanceFlags = Arrays.stream(flags.getUserFlags())
+            .filter(EVENT_ATTENDANCE_FLAGS::contains)
+            .flatMap(flag -> AttendanceStatus.fromUserFlag(flag).stream())
+            .toList();
+
+        if (eventAttendanceFlags.size() > 1) {
+            LOGGER.info("Unexpected: Multiple event attendance flags found: {}", eventAttendanceFlags);
+        }
+
+        return eventAttendanceFlags.stream().findAny();
+    }
+
+    public static Flags getEventAttendanceFlags() {
+        return FlagsBuilder.builder()
+            .add(EVENT_ATTENDANCE_FLAGS.toArray(new String[0]))
+            .build();
+    }
+}
--- a/tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
+++ b/tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
@@ -1,0 +1,92 @@
+package com.linagora.tmail.james.jmap;
+
+import java.util.List;
+
+import jakarta.inject.Inject;
+import jakarta.mail.Flags;
+
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.MailboxSession;
+import org.apache.james.mailbox.MessageIdManager;
+import org.apache.james.mailbox.MessageManager;
+import org.apache.james.mailbox.SessionProvider;
+import org.apache.james.mailbox.model.FetchGroup;
+import org.apache.james.mailbox.model.MailboxId;
+import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.MessageResult;
+import org.reactivestreams.Publisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public class StandaloneEventAttendanceRepository implements EventAttendanceRepository {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StandaloneEventAttendanceRepository.class);
+
+    private final MessageIdManager messageIdManager;
+    private final SessionProvider sessionProvider;
+
+    @Inject
+    public StandaloneEventAttendanceRepository(MessageIdManager messageIdManager, SessionProvider sessionProvider) {
+        this.messageIdManager = messageIdManager;
+        this.sessionProvider = sessionProvider;
+    }
+
+    @Override
+    public Publisher<AttendanceStatus> getAttendanceStatus(Username username, MessageId messageId) {
+        LOGGER.trace("Getting attendance status for user '{}' and message '{}'", username, messageId);
+        MailboxSession systemMailboxSession = sessionProvider.createSystemSession(username);
+
+        return getFlags(messageId, systemMailboxSession)
+            .flatMap(userFlags -> Mono.justOrEmpty(AttendanceStatus.fromMessageFlags(userFlags)))
+            .switchIfEmpty(handleMissingEventAttendanceFlag(messageId));
+    }
+
+    private Mono<AttendanceStatus> handleMissingEventAttendanceFlag(MessageId messageId) {
+        LOGGER.debug("""
+                No event attendance flag found for message {}.
+                Defaulting to NeedsAction
+                """, messageId);
+        return Mono.just(AttendanceStatus.NeedsAction);
+    }
+
+    @Override
+    public Publisher<Void> setAttendanceStatus(Username username, MessageId messageId,
+                                               AttendanceStatus attendanceStatus) {
+        MailboxSession systemMailboxSession = sessionProvider.createSystemSession(username);
+
+        return Flux.from(messageIdManager.getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, systemMailboxSession))
+            .map(MessageResult::getMailboxId)
+            .collectList()
+            .flatMap(mailboxIds ->
+                updateEventAttendanceFlagsInMailboxes(messageId, attendanceStatus, systemMailboxSession, mailboxIds));
+    }
+
+    private Mono<Void> updateEventAttendanceFlagsInMailboxes(MessageId messageId, AttendanceStatus attendanceStatus,
+                                                             MailboxSession session, List<MailboxId> mailboxIds) {
+        // By removing the current event attendance flags (without the new flag to set) we ensure
+        // commutativity of operations that is it does not matter whether the REMOVE or ADD operation is done first.
+        Flags eventAttendanceFlagsToRemove = AttendanceStatus.getEventAttendanceFlags();
+        eventAttendanceFlagsToRemove.remove(attendanceStatus.getUserFlag());
+        return Flux.concat(
+            messageIdManager.setFlagsReactive(
+                eventAttendanceFlagsToRemove,
+                MessageManager.FlagsUpdateMode.REMOVE,
+                messageId,
+                mailboxIds,
+                session),
+            messageIdManager.setFlagsReactive(
+                new Flags(attendanceStatus.getUserFlag()),
+                MessageManager.FlagsUpdateMode.ADD,
+                messageId,
+                mailboxIds,
+                session)
+        ).then();
+    }
+
+    private Flux<Flags> getFlags(MessageId messageId, MailboxSession session) {
+        return Flux.from(messageIdManager.getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, session))
+            .map(MessageResult::getFlags);
+    }
+}
--- a/tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepositoryTest.java
+++ b/tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepositoryTest.java
@@ -1,0 +1,205 @@
+package com.linagora.tmail.james.jmap;
+
+import java.util.List;
+import reactor.core.publisher.Mono;
+
+import static org.apache.james.mailbox.fixture.MailboxFixture.ALICE;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import jakarta.mail.Flags;
+
+import org.apache.james.mailbox.MailboxSession;
+import org.apache.james.mailbox.MailboxSessionUtil;
+import org.apache.james.mailbox.MessageUid;
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.mailbox.fixture.MailboxFixture;
+import org.apache.james.mailbox.inmemory.InMemoryMessageId;
+import org.apache.james.mailbox.inmemory.manager.InMemoryIntegrationResources;
+import org.apache.james.mailbox.model.FetchGroup;
+import org.apache.james.mailbox.model.Mailbox;
+import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.MessageResult;
+import org.apache.james.mailbox.store.MessageIdManagerTestSystem;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class StandaloneEventAttendanceRepositoryTest {
+    MailboxSession session;
+    StandaloneEventAttendanceRepository testee;
+    MessageIdManagerTestSystem messageIdManagerTestSystem;
+    Mailbox mailbox;
+
+    @BeforeEach
+    void setUp() throws MailboxException {
+        messageIdManagerTestSystem = createTestSystem();
+
+        testee = new StandaloneEventAttendanceRepository(messageIdManagerTestSystem.getMessageIdManager(),
+            messageIdManagerTestSystem.getMailboxManager().getSessionProvider());
+
+        session = MailboxSessionUtil.create(ALICE);
+        mailbox = messageIdManagerTestSystem.createMailbox(MailboxFixture.INBOX_ALICE, session);
+    }
+
+    @Test
+    void givenAcceptedFlagIsLinkedToMailGetAttendanceStatusShouldReturnAccepted() {
+        Flags flags = new Flags("$accepted");
+        MessageId messageId = createMessage(flags);
+        assertThat(Mono.from(testee.getAttendanceStatus(mailbox.getUser(), messageId)).block())
+            .isEqualTo(AttendanceStatus.Accepted);
+    }
+
+    @Test
+    void givenRejectedFlagIsLinkedToMailGetAttendanceStatusShouldReturnDeclined() {
+        Flags flags = new Flags("$rejected");
+        MessageId messageId = createMessage(flags);
+        assertThat(Mono.from(testee.getAttendanceStatus(mailbox.getUser(), messageId)).block())
+            .isEqualTo(AttendanceStatus.Declined);
+    }
+
+    @Test
+    void givenTentativelyAcceptedFlagIsLinkedToMailGetAttendanceStatusShouldReturnTentative() {
+        Flags flags = new Flags("$tentativelyaccepted");
+        MessageId messageId = createMessage(flags);
+        assertThat(Mono.from(testee.getAttendanceStatus(mailbox.getUser(), messageId)).block())
+            .isEqualTo(AttendanceStatus.Tentative);
+    }
+
+    // It should also print a warning message
+    @Test
+    void givenMoreThanEventAttendanceFlagIsLinkedToMailGetAttendanceStatusShouldReturnAny() {
+        Flags flags = new Flags("$rejected");
+        flags.add("$accepted");
+
+        MessageId messageId = createMessage(flags);
+        assertThat(Mono.from(testee.getAttendanceStatus(mailbox.getUser(), messageId)).block())
+            .satisfiesAnyOf(
+                status -> assertThat(status).isEqualTo(AttendanceStatus.Accepted),
+                status -> assertThat(status).isEqualTo(AttendanceStatus.Declined));
+    }
+
+    @Test
+    void getAttendanceStatusShouldFallbackToNeedsActionWhenNoFlagIsLinkedToMail() {
+        Flags flags = new Flags();
+        MessageId messageId = createMessage(flags);
+
+        assertThat(Mono.from(testee.getAttendanceStatus(mailbox.getUser(), messageId)).block())
+            .isEqualTo(AttendanceStatus.NeedsAction);
+    }
+
+    @Test
+    void getAttendanceStatusShouldFallbackToNeedsActionWhenNoEventAttendanceFlagIsLinkedToMail() {
+        Flags flags = new Flags(Flags.Flag.RECENT);
+        MessageId messageId = createMessage(flags);
+
+        assertThat(Mono.from(testee.getAttendanceStatus(mailbox.getUser(), messageId)).block())
+            .isEqualTo(AttendanceStatus.NeedsAction);
+    }
+
+    @Test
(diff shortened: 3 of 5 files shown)


chibenwa:
To be reviewed with @florentos17 first.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
```diff
@@ -0,0 +1,5 @@
+package com.linagora.tmail.james.jmap;
+
+public enum AttendanceStatus {
+    Accepted, Tentative, Delegated, NeedsAction, Declined
```
florentos17, line 4: what do you need the `Delegated` status for ? (not used anywhere)

florentos17 (commented):
looks good to me, it is straightforward

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
HoussemNasri: Delegated is declared in the [specification](https://datatracker.ietf.org/doc/html/rfc5545#section-3.2.12), but it is not in use for now, neither in the frontend nor the backend.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,124 @@
...
+        return getFlags(messageId, systemMailboxSession)
+            .map(this::getAttendanceStatusFromFlags);
+    }
+
+    @Override
+    public Publisher<Void> setAttendanceStatus(Username username, MessageId messageId,
+                                               AttendanceStatus attendanceStatus) {
+        MailboxSession systemMailboxSession = sessionProvider.createSystemSession(username);
+
+        return getFlags(messageId, systemMailboxSession)
```
chibenwa, line 53: We are getting flags more than twice before REPLACing them.

Two things about it:
 - it is innefficient
 - And get-then-set do not commute so it will be not concurrent friendly

I prefer us to add the one flag that interest us and remove the two others that do not interest us.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,124 @@
...
+        if (flags.contains("$accepted")) {
+            return AttendanceStatus.Accepted;
+        } else if (flags.contains("$rejected")) {
+            return AttendanceStatus.Declined;
+        } else if (flags.contains("$tentativelyaccepted")) {
+            return AttendanceStatus.Tentative;
+        } else {
+            return AttendanceStatus.NeedsAction;
+        }
+    }
```
chibenwa, line 103: I am pretty sure I can simplify this to the extreme!

I'd associate a flag string directly to the `AttendanceStatus` and then use this association to transform flags into AttendanceStatus and just return the first one.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        if (flags.contains("$accepted")) {
+            return AttendanceStatus.Accepted;
+        } else if (flags.contains("$rejected")) {
+            return AttendanceStatus.Declined;
+        } else if (flags.contains("$tentativelyaccepted")) {
+            return AttendanceStatus.Tentative;
+        } else {
+            return AttendanceStatus.NeedsAction;
```
chibenwa, line 101: This is implicit so the day we add an entry into EVENT_ATTENDANCE_FLAGS then we will return NeedsAction which would be a lie.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
HoussemNasri: Flags are implementation details and AttendanceStatus is an API. Remember we are going to have another implementation that uses DAV server.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    private AttendanceStatus getAttendanceStatusFromFlags(Flags flags) {
+        long eventAttendanceFlagsCount = Arrays.stream(flags.getUserFlags())
+            .filter(EVENT_ATTENDANCE_FLAGS::contains)
+            .count();
+
+        if (eventAttendanceFlagsCount > 1) {
+            LOGGER.warn("A message should not have more than one event attendance flag");
+            return AttendanceStatus.NeedsAction;
```
chibenwa, line 91: So `$accepted`, `$rejected` results in `$needs-action`? Weird!

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    private Mono<Flags> doSetFlags(MessageId messageId, Flags flagsToSet, MailboxSession session) {
+        try {
+            List<MailboxId> mailboxIds =
+                messageIdManager.getMessage(messageId, FetchGroup.MINIMAL, session)
+                    .stream().map(MessageResult::getMailboxId).toList();
+            messageIdManager.setFlags(flagsToSet, MessageManager.FlagsUpdateMode.REPLACE, messageId, mailboxIds,
+                session);
```
chibenwa, line 111:
```suggestion
            List<MailboxId> mailboxIds =
                messageIdManager.getMessage(messageId, FetchGroup.MINIMAL, session)
                    .stream().map(MessageResult::getMailboxId).toList();
            messageIdManager.setFlags(flagsToSet, MessageManager.FlagsUpdateMode.REPLACE, messageId, mailboxIds,
                session);
```

Please be reactive!

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
HoussemNasri: note: That would require that we introduce a new flag that does not exist on tmail-flutter.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
HoussemNasri: We can not tell which is the first because Flags uses a hashtable internally. The defined behavior in tests should be to return any of them.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
chibenwa: > Flags are implementation details

Not really it could be handy for some sort of retro-compatibility with today frontend behaviour.

And even without this this is more a remark about having code clean.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
florentos17: is there a use case for having several flags ? if not, setting a flag could replace the existing one (if any) so that only the most up to date flag remains ?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
chibenwa, line 4:
```suggestion
    Accepted("$accepted"), 
    Tentative("$tentative"), 
    Delegated("$delegated"),
    NeedsAction($needs-action"), 
    Declined("$declined");
    
    private final String userFlag;
    
    public static AttendanceStatus fromUserFlag(String userFlag) {
        // ...
    }
```

And get rid of the aweful map

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,129 @@
...
+        return getFlags(messageId, systemMailboxSession)
+            .flatMap(this::getAttendanceStatusFromFlags)
+            .switchIfEmpty(handleMissingEventAttendanceFlag(messageId, systemMailboxSession));
+    }
+
+    private Mono<AttendanceStatus> handleMissingEventAttendanceFlag(MessageId messageId, MailboxSession systemMailboxSession) {
+        return Mono.from(getFlags(messageId, systemMailboxSession)).map(flags -> {
+            LOGGER.warn("No event attendance flag found for message {}", messageId);
+            LOGGER.warn("Flags for message {}: {}", messageId, flags);
+            LOGGER.warn("Defaulting to NeedsAction for message {}", messageId);
```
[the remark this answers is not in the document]
chibenwa: BTW what we need getlag for? If it's just add it into the debug log it is not worth the extra DB read...

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,129 @@
...
+                        LOGGER.warn("No flag found for attendance status {}", attendanceStatus);
+                        LOGGER.warn("Falling back to $needs-action");
+                        return "$needs-action";
+                    });
+                flags.add(attendanceFlag);
+                return doSetFlags(messageId, flags, systemMailboxSession);
+            }).then();
+    }
+
+    private Optional<String> getFlagFromAttendanceStatus(AttendanceStatus attendanceStatus) {
```
chibenwa, line 74: This should be embedded into the type. See earlier comment

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private Mono<AttendanceStatus> getAttendanceStatusFromFlags(Flags flags) {
+        long eventAttendanceFlagsCount = Arrays.stream(flags.getUserFlags())
+            .filter(EVENT_ATTENDANCE_FLAGS::contains)
+            .count();
+
+        if (eventAttendanceFlagsCount > 1) {
+            LOGGER.warn("A message should not have more than one event attendance flag");
+        }
+
```
chibenwa, line 92:
```
Arrays.stream(flags.getUserFlsags())
    .flatMap(AttendanceStatus::fromUserFlag)
    .findAny()
    .orElse(NeedsAction)
```

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,129 @@
...
+        };
+    }
+
+    private Mono<AttendanceStatus> getAttendanceStatusFromFlags(Flags flags) {
+        long eventAttendanceFlagsCount = Arrays.stream(flags.getUserFlags())
+            .filter(EVENT_ATTENDANCE_FLAGS::contains)
+            .count();
+
+        if (eventAttendanceFlagsCount > 1) {
+            LOGGER.warn("A message should not have more than one event attendance flag");
```
chibenwa, line 90: Which message is that? Could be useful to know no?

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private Mono<Void> doSetFlags(MessageId messageId, Flags flagsToSet, MailboxSession session) {
+        return Mono.from(messageIdManager.getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, session))
+            .map(MessageResult::getMailboxId)
+            .flux()
+            .collectList()
+            .flatMap(mailboxIds ->
+                Mono.from(messageIdManager.setFlagsReactive(
+                    flagsToSet,
+                    MessageManager.FlagsUpdateMode.REPLACE,
```
chibenwa, line 114: That's the point: I do not want get-then-REPLACE

I want 2 calls to set flags: Add one flag and remove two.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
HoussemNasri: > setting a flag could replace the existing one (if any) so that only the most up to date flag remains ?

True, this is implemented in the setAttendanceStatus but just in case for some unknown reason ( e.g. concurrency issue) we have found that multiple event attendance flags are associated with the same message, we return any of them and log a warning.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,129 @@
...
+                    messageId,
+                    mailboxIds,
+                    session
+                ))
+            )
```
chibenwa, line 119:
```suggestion
                    session)))
```

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,129 @@
...
+            .onErrorContinue((error, object) ->
+                LOGGER.error("Error while setting flags for message ID {}", messageId, error)
+            );
```
chibenwa, line 122:
```suggestion
```

We are not handling the error, the log gonna be duplicated. Better remove!

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepositoryTest.java
```diff
@@ -0,0 +1,180 @@
...
+            .block();
+    }
+
+    private MessageId createMessage(Flags flags) {
+        MessageId messageId = messageIdManagerTestSystem.persist(
+            mailbox.getMailboxId(),
+            MessageUid.of(111),
+            flags,
+            session
+        );
```
chibenwa, line 157:
```suggestion
            session);
```

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepositoryTest.java
```diff
@@ -0,0 +1,180 @@
...
+    void setAttendanceStatusShouldSetAcceptedFlag() {
+        Flags flags = new Flags();
+        MessageId messageId = createMessage(flags);
+
+        Mono.from(testee.setAttendanceStatus(mailbox.getUser(), messageId, AttendanceStatus.Accepted)).block();
+
+        Flags updatedFlags = getFlags(messageId);
+
+        assertThat(updatedFlags.contains("$accepted")).isTrue();
+    }
```
chibenwa, line 108: Let's test idempotance:

```suggestion
    @Test
    void setAttendanceStatusShouldSetAcceptedFlag() {
        Flags flags = new Flags();
        MessageId messageId = createMessage(flags);

        Mono.from(testee.setAttendanceStatus(mailbox.getUser(), messageId, AttendanceStatus.Accepted)).block();
        Mono.from(testee.setAttendanceStatus(mailbox.getUser(), messageId, AttendanceStatus.Accepted)).block();

        Flags updatedFlags = getFlags(messageId);

        assertThat(updatedFlags.contains("$accepted")).isTrue();
        assertThat(updatedFlags.length).isEqualTo(1);
    }
```

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepositoryTest.java
```diff
@@ -0,0 +1,180 @@
...
+        Flags flags = new Flags("$accepted");
+        MessageId messageId = createMessage(flags);
+
+        Mono.from(testee.setAttendanceStatus(mailbox.getUser(), messageId, AttendanceStatus.Declined)).block();
+
+        Flags updatedFlags = getFlags(messageId);
+        assertThat(updatedFlags.contains("$accepted")).isFalse();
+        assertThat(updatedFlags.contains("$rejected")).isTrue();
+    }
+
```
chibenwa, line 143: Let's also try to set needsAction

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
HoussemNasri: Delegated is not in use now. It would additionally require some extra handling because, in addition to the delegated flag, we need to keep track of whom we are delegating our event attendance status, so I'm thinking of just removing it from the enum.

tmail-backend/jmap/extensions/src/test/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepositoryTest.java
```diff
@@ -0,0 +1,193 @@
...
+        assertThat(updatedFlags.contains("$accepted")).isTrue();
+        assertThat(updatedFlags.getUserFlags().length).isEqualTo(1);
+    }
+
+    private Flags getFlags(MessageId messageId) {
+        return Mono.from(messageIdManagerTestSystem.getMessageIdManager()
+                .getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, session))
+            .map(MessageResult::getFlags)
+            .block();
+    }
```
vttranlina, line 163: We can add one more test for:
- Given: message A with Accepted
- When: update message A with Rejected
- Then: attendee flags of message A contain only Rejected

HoussemNasri: Such test already exist: setAttendanceStatusShouldRemoveExistingEventAttendanceFlags

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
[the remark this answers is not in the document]
HoussemNasri: This log will be triggered when no attendance flags exist for that message. We log the flags in case an attendance flag was attached to the message, but we failed to recognize that.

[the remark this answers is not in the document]
HoussemNasri: Downgraded the log level to debug.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
```diff
@@ -0,0 +1,67 @@
...
+
+    AttendanceStatus(String userFlag) {
+        this.userFlag = userFlag;
+    }
+
+    public String getUserFlag() {
+        return userFlag;
+    }
+
+    public static Optional<AttendanceStatus> fromUseFlag(String userFlag) {
```
chibenwa, line 39:
```suggestion
    public static Optional<AttendanceStatus> fromUserFlag(String userFlag) {
```

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
```diff
@@ -0,0 +1,67 @@
...
+        Preconditions.checkNotNull(userFlag);
+        for (AttendanceStatus status : AttendanceStatus.values()) {
+            if (status.userFlag.equals(userFlag)) {
+                return Optional.of(status);
+            }
+        }
```
chibenwa, line 45: Arrays.stream(AttendanceStatus.values())...

Be functional

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/AttendanceStatus.java
```diff
@@ -0,0 +1,67 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+
+public enum AttendanceStatus {
+    Accepted("$accepted"),
+    Declined("$rejected"),
+    Tentative("$tentativelyaccepted"),
+    NeedsAction("$needs-action");
```
chibenwa, line 21: We removed the spec-defined $Delegated ? 

Why?

HoussemNasri: Delegated is not in use now. It would additionally require some extra handling because, in addition to the delegated flag, we need to keep track of whom we are delegating our event attendance status, so I'm thinking of just removing it from the enum.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,90 @@
...
+                                               AttendanceStatus attendanceStatus) {
+        MailboxSession systemMailboxSession = sessionProvider.createSystemSession(username);
+
+        return updateEventAttendanceFlags(messageId, attendanceStatus, systemMailboxSession);
+    }
+
+    private Mono<Void> updateEventAttendanceFlags(MessageId messageId, AttendanceStatus attendanceStatus, MailboxSession session) {
+        return Mono.from(messageIdManager.getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, session))
+            .map(MessageResult::getMailboxId)
+            .flux()
```
chibenwa, line 67:
```suggestion
        return Flux.from(messageIdManager.getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, session))
            .map(MessageResult::getMailboxId)
```

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,90 @@
...
+                        MessageManager.FlagsUpdateMode.REMOVE,
+                        messageId,
+                        mailboxIds,
+                        session),
+                    messageIdManager.setFlagsReactive(
+                        new Flags(attendanceStatus.getUserFlag()),
+                        MessageManager.FlagsUpdateMode.ADD,
+                        messageId,
+                        mailboxIds,
+                        session)
```
chibenwa, line 82: `REMOVE(A, B, C)` do not commute with `ADD(A)`

`REMOVE(B, C)` do commute with `ADD(A)`

So please manage NOT to unset in the remove the one flag we are about to add.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -0,0 +1,90 @@
...
+                ).then());
+    }
+
+    private Mono<Flags> getFlags(MessageId messageId, MailboxSession session) {
+        return Mono.from(messageIdManager.getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, session))
```
chibenwa, line 87:
```suggestion
    private Flux<Flags> getFlags(MessageId messageId, MailboxSession session) {
        return Flux.from(messageIdManager.getMessagesReactive(List.of(messageId), FetchGroup.MINIMAL, session))
```

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/StandaloneEventAttendanceRepository.java
```diff
@@ -55,28 +56,33 @@ public Publisher<Void> setAttendanceStatus(Username username, MessageId messageI
...
+                messageId,
+                mailboxIds,
+                session),
+            messageIdManager.setFlagsReactive(
+                new Flags(attendanceStatus.getUserFlag()),
+                MessageManager.FlagsUpdateMode.ADD,
+                messageId,
+                mailboxIds,
+                session)
+        ).then();
```
chibenwa, line 85:
```suggestion
                session))
            .then();
```

chibenwa (approved):

chibenwa:
OK as it stands but in the Standalone version, I believe it is the EventAttendance repository that actually must encapsulate "sending a mail back to the organizer" which is currently unwired.

We would need to implement this in a follow up PR...

vttranlina (approved):

Arsnael:
Can we have:
- cleanup of git commits?
- issue number in the commit descriptions?

HoussemNasri:
> OK as it stands but in the Standalone version, I believe it is the EventAttendance repository that actually must encapsulate "sending a mail back to the organizer" which is currently unwired.

https://github.com/linagora/tmail-backend/issues/1400

Arsnael (approved):

PR #1406, opened 2024-12-11
Move memory implementations in jmap-extensions to jmap-extensions-api

### LINKED ISSUE #1399: Relocate memory implementations from jmap-extensions to jmap-extensions-api
All memory implementations (and tests) in `jmap-extensions` should be relocated to `jmap-extensions-api`.

## definition of done
All `Memory*` prefixed classes reside in `jmap-extensions-api`.

*Originally posted: https://github.com/linagora/tmail-backend/pull/1339#discussion_r1875941249*

HoussemNasri:
Closes #1399


tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/firebase/MemoryFirebaseSubscriptionRepositoryModule.java
```diff
@@ -0,0 +1,19 @@
+package com.linagora.tmail.james.jmap.firebase;
+
+import org.apache.james.user.api.DeleteUserDataTaskStep;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Scopes;
+import com.google.inject.multibindings.Multibinder;
+
+public class MemoryFirebaseSubscriptionRepositoryModule extends AbstractModule {
```
chibenwa, line 9: I feel like memory guice modules...

 - Are not realy related to the transport/protocol layer
 - Are only used by apps/memory-app
 
 Why not relocate them to apps/memory-app

HoussemNasri: Nice idea!

tmail-backend/apps/memory/pom.xml
```diff
@@ -179,6 +179,17 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>net.alchim31.maven</groupId>
+                <artifactId>scala-maven-plugin</artifactId>
```
chibenwa, line 184: Sorry for bothering but things like guice modules do not need to be scala.

It adds build delay.

Is it possible to rewrite those in java and drop this plugin?

chibenwa (approved):

PR #1412, opened 2024-12-16
[ENHANCEMENT] Make the Apisix plugin reactive

chibenwa:
Please:

 - [ ] Test this first on a real Apisix
 - [ ] Regular JMAP performance tests before / after
 - [ ] Conduct performance tests without Redis
 
 CF https://github.com/linagora/tmail-backend/issues/1247
 
 @vttranlina @hungphan227 please cary over this work from now on

### DIFF
--- a/demo/apisix/tmail-apisix-plugin-runner/src/main/java/com/linagora/apisix/plugin/AppConfiguration.java
+++ b/demo/apisix/tmail-apisix-plugin-runner/src/main/java/com/linagora/apisix/plugin/AppConfiguration.java
@@ -16,8 +16,10 @@ import org.springframework.util.StringUtils;
 import io.lettuce.core.ReadFrom;
 import io.lettuce.core.RedisClient;
 import io.lettuce.core.RedisURI;
-import io.lettuce.core.api.sync.RedisStringCommands;
+import io.lettuce.core.api.reactive.RedisReactiveCommands;
+import io.lettuce.core.api.reactive.RedisStringReactiveCommands;
 import io.lettuce.core.cluster.RedisClusterClient;
+import io.lettuce.core.cluster.api.reactive.RedisAdvancedClusterReactiveCommands;
 import io.lettuce.core.codec.StringCodec;
 import io.lettuce.core.masterreplica.MasterReplica;
 import io.lettuce.core.masterreplica.StatefulRedisMasterReplicaConnection;
@@ -56,7 +58,7 @@ public class AppConfiguration {
         return new IRevokedTokenRepository.MemoryRevokedTokenRepository();
     }
 
-    private RedisStringCommands<String, String> initRedisCommand() {
+    private RedisStringReactiveCommands<String, String> initRedisCommand() {
         Duration timeoutDuration = Duration.ofMillis(redisTimeout);
 
         logger.info("The plugin using redis {} for storage revoked tokens.\n" +
@@ -76,14 +78,14 @@ public class AppConfiguration {
         }
     }
 
-    public static RedisStringCommands<String, String> initRedisCommandSentinel(String redisUrl,
+    public static RedisReactiveCommands<String, String> initRedisCommandSentinel(String redisUrl,
                                                                                Duration redisTimeout) {
         RedisURI redisURI = RedisURI.create(redisUrl);
         redisURI.setTimeout(redisTimeout);
-        return RedisClient.create(redisURI).connect().sync();
+        return RedisClient.create(redisURI).connect().reactive();
     }
 
-    public static RedisStringCommands<String, String> initRedisCommandStandalone(String redisUrl,
+    public static RedisReactiveCommands<String, String> initRedisCommandStandalone(String redisUrl,
                                                                                  String redisPassword,
                                                                                  Duration redisTimeout) {
         String[] redisUrlParts = redisUrl.split(":");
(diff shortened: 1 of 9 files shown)


demo/apisix/tmail-apisix-plugin-runner/src/main/java/com/linagora/apisix/plugin/TokenRevokedFilter.java
```diff
@@ -29,25 +31,32 @@ public String name() {
     @Override
     public void filter(HttpRequest request, HttpResponse response, PluginFilterChain chain) {
         logger.debug("Received a new request");
-        sidFilter(request, response);
-        chain.filter(request, response);
+        sidFilter(request, response)
```
vttranlina, line 34: It's possible that `chain.filter(request, response)` will be called AFTER (bad case) or BEFORE (perfect case) the "PluginFilterChain" had completed

+1 :beer: if it works

chibenwa: It's callback based, it is fine!
That's Netty under the hood it feels like home.

demo/apisix/tmail-apisix-plugin-runner/src/main/java/com/linagora/apisix/plugin/AppConfiguration.java
```diff
@@ -39,7 +39,7 @@ public class AppConfiguration {
     @Value("${redis.password}")
     private String redisPassword;
 
-    @Value("${redis.timeout:5000}")
+    @Value("${redis.timeout:500}")
```
quantranhong1999, line 42: +1 for this

demo/apisix/tmail-apisix-plugin-runner/src/main/java/com/linagora/apisix/plugin/TokenRevokedFilter.java
hungphan227: > It's callback based, it is fine! That's Netty under the hood it feels like home.

Could you refer to any document that prove this "callback"? I agree with Tung. Code we put in reactor could be executed after response is sent to client.

chibenwa: You are tiring me, so deeply, some days.

chibenwa: Because implementing a callback interface:

 - > it's clear that though asynchronous the callbacks are still executed in order
 - > And sending the response to the client would trivially be implemented in the very last filter.

hungphan227: I have tried running apisix with this new reactive code and saw that when redis down, client immediately received response after sending jmap request although I set redis timeout more than 10s ----> I don't think there is any "callback" here.

chibenwa: Isn't it even better?

Driver might detect "it's down" and not send the request in the first place?

Can we validate logout works?

hungphan227: I have tried again with redis up. I revoked the token but client still received code 200 after sending jmap request (it is supposed to be 401).

chibenwa: Chain works

And is compatible with async code

image

hungphan227: When the "subscribe" method in TokenRevokedFilter class is executed, APISIX understands that all filters has completed and returns the response to client before running reactive code.

hungphan227: I have written a test to prove that the client could receive response before reactive code is executed.

https://github.com/hungphan227/tmail-backend/commit/5fe214c11d736c9c49b46f0a2b687dc66002ad5b

The above test pass with the old code but fail with the reactive code.

chibenwa: THen let's patch APisix!

chibenwa: https://github.com/apache/apisix-java-plugin-runner/pull/313

Please qualify the new plugin with the new plugin runner ;-)

hungphan227: Yeah with the updated plugin runner, the reactive code works well.

I have tested case token revoked --> pass
I have conducted performance test like this https://github.com/linagora/tmail-backend/issues/1247#issuecomment-2545209866 --> The result is good. The latency is always around 5s (I set timeout to 5s).

Regarding updating the plugin runner, I wonder if there is anyone reviewing and approving the PR. In case the repository is no longer maintained, I suggest we fork, create our own repository for the plugin runner (as well as CI/CD)

chibenwa: > Regarding updating the plugin runner, I wonder if there is anyone reviewing and approving the PR.

It is Christmass time.

Be patient.

We can package, build and deploy the PR for the time being.

Can we do a JMAP perf test on top of this change?
Can we publish the updated Apisix docker image with both changes to the runner and to the plugin?

hungphan227: Apisix docker image: hungphan227/apisix:3.9.1-debian-javaplugin

hungphan227: It is complicated to update staging env to run jmap perf test with apisix. Therefore, I will update james-gatling and run it locally with tmail-backend memory version.

hungphan227:
Performance test result with 2 min inject, 8 min scenario, 1000 users, tmail memory (due to an unknown issue, I have to remove some JMAP requests in the test):

- Regular JMAP performance tests before new plugin:
Screenshot from 2025-01-02 18-08-53


- JMAP performance tests without redis before new plugin:
Screenshot from 2025-01-02 22-36-07


- Regular JMAP performance tests with new plugin:
Screenshot from 2025-01-02 20-24-27


- JMAP performance tests without redis with new plugin:
Screenshot from 2025-01-02 21-28-54

chibenwa:
Ok thanks a lot @hungphan227 !

This proves:
 - 1/ Small gains on regular operations (5% mean time AND 7.6 % P99)
 - 2/ Massive gains in term of ourage (x15 mean time AND  x50 P99)

Which image did you use to run those tests?

Arsnael (commented):
LGTM. is this still a draft? or can we switch it to ready for review?

hungphan227:
> Ok thanks a lot @hungphan227 !
> 
> This proves:
> 
>     * 1/ Small gains on regular operations (5% mean time AND 7.6 % P99)
> 
>     * 2/ Massive gains in term of ourage (x15 mean time AND  x50 P99)
> 
> 
> Which image did you use to run those tests?

Apisix docker image with new plugin: hungphan227/apisix:3.9.1-debian-javaplugin

chibenwa:
> LGTM. is this still a draft? or can we switch it to ready for review?

Build is not necessarily straight forward as it requires building the java plugin yourself.

Maybe keep this as a draft until upstream is released?

chibenwa:
Code is merged upstream, I am waiting for the plugin runner upgrade in order to update this.

PR #1436, opened 2025-01-02
[FIX] Inject OpenPaaS module correctly

chibenwa:
First bug of 2025 ! :-)

### DIFF
--- a/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/configuration/OpenPaasConfiguration.java
+++ b/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/configuration/OpenPaasConfiguration.java
@@ -4,7 +4,6 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
 
 import org.apache.commons.configuration2.Configuration;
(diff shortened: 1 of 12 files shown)


Arsnael (approved):

Arsnael:
@chibenwa it's not the first time lately I fix wrong imports order on your PRs, can you please check your IDE configuration regarding imports order? Thanks :)

quantranhong1999 (approved):

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/configuration/OpenPaasConfiguration.java
```diff
@@ -57,14 +60,17 @@ public static boolean isConfiguredContactConsumer(Configuration configuration) {
         return readRabbitMqUri(configuration).isPresent();
     }
 
-    public static Optional<AmqpUri> readRabbitMqUri(Configuration configuration) {
+    public static Optional<List<AmqpUri>> readRabbitMqUri(Configuration configuration) {
```
HoussemNasri, line 63: We don't need the Optional wrapper here. An empty list should do.

chibenwa: Feel free to open a follow up PR to clean this up.

PR #1446, opened 2025-01-08
ISSUE-1445 LDAP mailing lists must preserve original Reply To

chibenwa:


hoangdat:
I still not really understand it. But is it still ok for this case below: original sender dont really want to receive reply, but `reply-to address` need it? 

signal-2025-01-09-101025_002

PR #1452, opened 2025-01-10
Disable AIBotMailetTest

quantranhong1999:
Somehow OpenAI returns the `insufficient_quota` error (even on my local machine). Could be something hiding behind.

Let's disable this test for now.


quantranhong1999:
Duplicated https://github.com/linagora/tmail-backend/pull/1451

PR #1460, opened 2025-01-14
ISSUE-1299 Fix DistributedLinagoraSecondaryBlobStoreTest and TmailEventSerializer

Arsnael:
When putting back the distributed integration tests for secondary blobstore, it was broken. I think tmaileventserializer is used by the secondary blobstore, and does not support multiple event serialization. After fixing this the tests are green.

@chibenwa what version of tmail running on cnb prod? could the secondary blobstore been affected lately?


chibenwa (approved):

chibenwa:
> could the secondary blobstore been affected lately?

Indeed. good catch.

> @chibenwa what version of tmail running on cnb prod?

An older backport.

PR #1463, opened 2025-01-15
[ENHANCEMENT] Cache webfinger call for 1 day

chibenwa:


Arsnael (approved):

quantranhong1999 (approved):

guimard:
1 day isn't a good value, prefer "little bit more than one day", ie 30 hours, or best 80 hours so that no problem on Monday morning, what do you think ? Maybe a local parameter ?

PR #1466, opened 2025-01-15
Magic attachments

chibenwa:
Very WIP

I need to:
 - [ ] Triage those changes locally
 - [ ] Needs "benchmarks" - gatling?
 - [ ] And indeed to reduce the boiler plate eg by not copying `TMailCleverMailboxSessionMapperFactory` which makes for halth of the code base added here
 
 See the related James PR


chibenwa:
Also it would need code reorganisation so that reusable pieces can benefit the PG app, for instance...

chibenwa:
Needs https://github.com/apache/james-project/pull/2622

Local simple perf tests results: a message with 60 attachments.

Before downloading one took ~100 ms
After downloading one took ~40 ms

(knowing this shows only part of the improvment as on the testing S3 data is held in memory)

chibenwa:
TODO 

 - [ ] Relocate the JMAP code in JMAP packages
 - [ ] And I bet merge ?
 - [ ] Do not forget to create a backport ticket for PG...

Arsnael (approved):

Arsnael:
Not sure why the github UI isn't updated but green on the CI. An other review perhaps?

chibenwa:
image

It turned out build is green

PR #1476, opened 2025-01-17
ISSUE-1299 [Build] Make DistributedLinagoraSecondaryBlobStoreTest mor…

Arsnael:
…e stable

By moving static init code to proper BeforeAll static method, on top of being more clean, seems to make the flaky test more stable.

Did try a couple of things, could always reproduce locally flaky test by repeating it 50 times (would start breaking around 20 or 30 runs). But with this, run twice 50 times and was green all the way


quantranhong1999 (approved):

vttranlina (approved):
+100000

I got CI fail on pr https://github.com/linagora/tmail-backend/pull/1470 that relate to it

hungphan227:
Nice. I couldn't reproduce in my local before

hungphan227 (approved):

PR #1554, opened 2025-02-19
Create calendar event invitations dynamically in LinagoraCalendarEventAcceptMethodContract

### LINKED ISSUE #1545: Restructure relevant JMAP tests to cover CalDavEventAttendanceRepository
## Why?

Because for CalDav we need to enable OpenPaas integration and not all tests are applicable for both Standalone and CalDav implementations.

## How?

Create: 
- `DavMemoryLinagoraCalendarEventAcceptMethodTest`
- `DavMemoryLinagoraCalendarEventRejectMethodTest`
- `DavMemoryLinagoraCalendarEventMaybeMethodTest`,
- `DavMemoryLinagoraCalendarEventAttendanceGetMethodTest` 

that launches an OpenPaas and CalDav server along the TMail server to run the integration tests on.

The implementation is based on DAVClient, which does not utilize storage. Therefore, tests with the Distributed Server are not necessary.

Tests specific to the Standalone or CalDav implementations should be placed in a dedicated contract test class.

We need to create a new user for each test to achieve isolation. This would require that we generate email invitations dynamically.

## DoD

- Tests for CalDavEventAttendanceRepository should run on a real/dockerized OpenPaas server.
- ~100% test coverage of CalDavEventAttendanceRepository

HoussemNasri:
Contributes to #1545


tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/pom.xml
```diff
@@ -94,6 +94,22 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>tmail-openpaas</artifactId>
+            <type>test-jar</type>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>tmail-openpaas</artifactId>
+            <scope>test</scope>
+        </dependency>
```
chibenwa, line 106: In a normal world this dependency is not needed

tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/src/test/java/com/linagora/tmail/james/DavMemoryLinagoraCalendarEventAcceptMethodTest.java
```diff
@@ -0,0 +1,179 @@
...
+import com.linagora.tmail.james.app.MemoryServer;
+import com.linagora.tmail.james.common.InvitationEmailData;
+import com.linagora.tmail.james.common.LinagoraCalendarEventAcceptMethodContract;
+import com.linagora.tmail.james.common.User;
+import com.linagora.tmail.james.jmap.EventAttendanceRepository;
+import com.linagora.tmail.james.jmap.firebase.FirebaseModuleChooserConfiguration;
+import com.linagora.tmail.module.LinagoraTestJMAPServerModule;
+
+
+public class DavMemoryLinagoraCalendarEventAcceptMethodTest implements
```
chibenwa, line 73: The entire point of a module chooser is to easily turn on/off modules without the kind of ceremony in this class.

tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/src/test/resources/mailetcontainer_openpaas_deployment.xml
```diff
@@ -0,0 +1,179 @@
+<?xml version="1.0"?>
```
chibenwa, line 1: Why we need a custom mailet container when OpenPaaS?
Is it *really* needed?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/dav/DavCalendarObject.java
```diff
@@ -57,11 +57,7 @@ public static Optional<DavCalendarObject> fromDavResponse(DavResponse davRespons
     }
 
     private static String normalizeVCalendarData(CalendarData calendarData) {
-        return calendarData.getValue()
-            .lines()
-            .map(String::trim)
-            .filter(Predicate.not(String::isBlank))
-            .collect(Collectors.joining("\n"));
+        return calendarData.getValue();
```
chibenwa, line 60: Rationals for this change?

tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/src/test/resources/mailetcontainer_openpaas_deployment.xml
HoussemNasri: Hmmm, I think so. We could have one `mailetcontainer.xml` that defines the `OpenPaasAmqpForwardAttribute` but that would lead to server failing when for tests that have OpenPaas disabled.

tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/src/test/java/com/linagora/tmail/james/DavMemoryLinagoraCalendarEventAcceptMethodTest.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private static Module provideDavUserProvider() {
+        return new AbstractModule() {
+            @Provides
+            public DavUserProvider provideDavUserProvider(OpenPaasRestClient openPaasClient) {
+                return new OpenPaasDavUserProvider(openPaasClient);
+            }
+        };
+    }
```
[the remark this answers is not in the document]
HoussemNasri: Okay, I am opening a separate pull request, don't want to bloat this one.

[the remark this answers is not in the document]
chibenwa: Then let's just postpone `DavMemoryLinagoraCalendarEvent_____MethodTest` so the guice modules are not needed and we actually only update the contract ?

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/LinagoraCalendarEventMethodContractUtilities.scala
```diff
@@ -18,16 +18,103 @@
...
+
+    val templateAsString = ClassLoaderUtils.getSystemResourceAsString(invitationEmailTemplate)
+
+    val emailTemplate = Mustache.compiler
+      .withLoader((name: String) => new InputStreamReader(ClassLoaderUtils.getSystemResourceAsSharedStream("template/" + name)))
+      .compile(templateAsString)
+
+    val mail = emailTemplate.execute(invitationEmailData)
+
+    println(mail)
```
chibenwa, line 79: debug spotted

tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/src/test/resources/mailetcontainer_openpaas_deployment.xml
chibenwa: As said in the call we need this because we need a RabbitMQ to push changes to.
We agreed (in a follow up PR)  torun those tests not on memory but on distributed where all distributed tests configures and set up this mailet.

PR #1569, opened 2025-02-26
Create calendar event invitations dynamically in LinagoraCalendarEventAcceptMethodContract

HoussemNasri:
Contributes to #1545


HoussemNasri:
I believe we can merge this now and propagate the changes later to the reject and maybe methods. By introducing DavMemoryLinagoraCalendarEvent, we would be making additional changes, and it would be cumbersome to duplicate the same changes that we will override later.

chibenwa (approved):

chibenwa:
(waiting the final green build but yes I do agree)

Arsnael (approved):

PR #1586, opened 2025-03-04
[ENHANCEMENT] Use sentinel for PubSub use case

chibenwa:
Using standalone for sure will prevent failover!


Arsnael:
The compilation fails => 

[ERROR] Failed to execute goal net.alchim31.maven:scala-maven-plugin:4.9.1:compile (scala-compile-first) on project tmail-event-bus-redis: Execution scala-compile-first of goal net.alchim31.maven:scala-maven-plugin:4.9.1:compile failed: Compilation failed: InterfaceCompileFailed -> [Help 1]

Might need to check the lettuce api for createSentinelClient

Arsnael:
Honestly, just pushing a quick fixup but... this does not change anything really. Because the createSentinelClient is pretty much a copy/paste of the createMasterReplicaClient in RedisClientFactory.

I noticed that in redis sentinel conf, there is a readFrom paramter, defaulting to MASTER node, that is not... being used in redis client setup. Need to double check on lettuce doc but could be maybe what we are missing?

I might open a ticket, I think this RedisClientFactory needs some reworking. Some other params could be used in some other topologies but aren't, like `ioThreads` and `workerThreads`, the cluster redis setup does not take ssl configuration into account, etc...

Arsnael:
=> https://github.com/linagora/james-project/issues/5412

PR #1618, opened 2025-03-17
ISSUE-1574 Create CalDavCollect mailet

hungphan227:


Arsnael (commented):
Read it nothing to add

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -0,0 +1,96 @@
...
+        } catch (ClassCastException e) {
+            LOGGER.error("Received a mail with {} not being an ICAL object for mail {}", sourceAttributeName, mail.getName(), e);
+        }
+    }
+
+    private void handleCalendarInMail(Calendar calendar, Mail mail) {
+        mail.getMaybeSender().asOptional()
+            .ifPresent(sender ->
+                davUserProvider.provide(Username.of(sender.asString()))
+                    .map(davUser -> davClient.getCalendarObject(davUser, getEventUid(calendar))
```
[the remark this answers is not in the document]
hungphan227: check if calendar object exist. if not proceed to create new

tmail-backend/integration-tests/jmap/distributed-jmap-integration-tests/src/test/resources/mailetcontainer.xml
```diff
@@ -77,17 +77,8 @@
                 <attribute>icalendar2</attribute>
                 <onMailetException>ignore</onMailetException>
             </mailet>
-            <mailet match="SenderIsLocal" class="org.apache.james.transport.mailets.ICALToJsonAttribute">
+            <mailet match="All" class="com.linagora.tmail.mailet.CalDavCollect">
```
chibenwa, line 80: Keep the matcher maybe?

com.linagora.tmail.mailet.IsOpenPaasSetup

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/CalDavCollectIntegrationTest.java
```diff
@@ -0,0 +1,163 @@
...
+                    return new OpenPaasConfiguration(
+                        openPaasServerExtension.getBaseUrl(),
+                        ALICE_ID,
+                        WireMockOpenPaaSServerExtension.GOOD_PASSWORD,
+                        false,
+                        davServerExtension.getDavConfiguration());
+                }
+
+                @ProvidesIntoSet
+                public BlobResolver provideFakeBlobResolver() {
```
chibenwa, line 106: :-1:

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -0,0 +1,121 @@
...
+            LOGGER.error("Received a mail with {} not being an ICAL object for mail {}", sourceAttributeName, mail.getName(), e);
+        }
+    }
+
+    private void handleCalendarInMail(Calendar calendar, Mail mail) {
+        try {
+            mail.getMaybeSender().asOptional()
+                .ifPresent(sender ->
+                    davUserProvider.provide(Username.of(sender.asString()))
+                        .flatMap(davUser -> davClient.getCalendarObject(davUser, getEventUid(calendar))
```
chibenwa, line 92: Again see my previous comment.

We **MUST** resolve the calendar home of the principal !

tmail-backend/integration-tests/jmap/distributed-jmap-integration-tests/src/test/resources/mailetcontainer.xml
Arsnael: Should be an other matcher then, as `IsOpenPaasRabbitMqSetup` is relying on `OpenPaasRabbitMQChannelPoolHolder` and we want to get rid of it at some point.

 I think we should as well mark `IsOpenPaasRabbitMqSetup` matcher and `OpenPaasAmqpForwardAttribute` as deprecated while we are at it

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
hungphan227:
```
public static final String DAV_ADMIN = "admin";
public static final String DAV_ADMIN_PASSWORD = "secret123";
```
call dav to get calendar home set will fail with this credentials (401 Unauthorized)

Arsnael: It works, sat down with @hungphan227 to prove it.

Please now implement it in the code :)

hungphan227: Thanks @Arsnael. Probably need to update this https://github.com/linagora/tmail-backend/blob/master/tmail-backend/tmail-third-party/openpaas/docker-compose-carddav-sample.yml (I failed to test the api with this)

hungphan227: Internally, davClient.getCalendarObject use PROFIND http://172.20.0.2/calendars/67dc1c9182ecf75dd94a1d97 to get:

`<d:href>/calendars/67dc1c9182ecf75dd94a1d97/67dc1c9182ecf75dd94a1d97/</d:href>`

Then it use the uri with eventUid in the body to get calendar object.

However, if using http://172.20.0.2/principals/users/67dc1c9182ecf75dd94a1d97, the result would be:

`<d:href>/calendars/67dc1c9182ecf75dd94a1d97/</d:href>`

Using above uri to get calendar object will return wrong result.

I think we should not change anything.

chibenwa: > Internally, davClient.getCalendarObject use PROFIND http://172.20.0.2/calendars/67dc1c9182ecf75dd94a1d97 to get 

PLEASE for the time number 3 implement exactly the principal call test here https://ci.linagora.com/btellier/dav-integrationtesting/-/blob/master/src/test/java/com/linagora/dav/PrincipalTest.java#L128

hungphan227: To summarize the idea:

- Step 1: Use api PROPFIND /principals/users/67dcd96831d07e1c3c27c091 to get calendar home set. The response look like:
```
<cal:calendar-home-set>
        <d:href>/calendars/67dcd96831d07e1c3c27c091/</d:href>
</cal:calendar-home-set>
```
- Step 2: Use the link in the result of step 1 (/calendars/67dcd96831d07e1c3c27c091) to get the list of calendar boxes. The response look like:
```
<d:response>
        <d:href>/calendars/67dcd96831d07e1c3c27c091/67dcd96831d07e1c3c27c091/</d:href>
</d:response>
```
- Step 3: Use the like in the result of step 2 to do things: get specific calendar, create new calendar

In the current code, the method davClient.getCalendarObject has already implemented step 2 and step 3. As I understand, step 1 is missing. Is that correct?

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/CalDavCollectIntegrationTest.java
hungphan227: what is the problem?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
chibenwa: The issue I have with your proposal is that it takes the risk of creating the event in ANY of the user calendar.

We want it to be the default one and sadly we have no real way apparently in DAV to clearly identify it.

Having had a quick look about it my impression is that we would anyway hardcode something OpenPaaS specific

So maybe we can simplify this a lot:

 - Resole the openPaaSID to OpenPaaS
 - Build the URL
 - and simply PUT into it
 
 (and completly forget this bad idea of principal, sorry)
 
 WDYT?

hungphan227: "Resolve the openPaaSID to OpenPaaS" ---> does this mean calling to openpass to get userId?

chibenwa: Yes

hungphan227: I have already done all those things (get userId, create URL, send PUT request to dav) in this PR. Regarding davClient.getCalendarObject, the reason I used it is to check if the calendar event has already existed. If not, proceed to creating new one. Otherwise do updating calendar event which probably will be done in other ticket.

tmail-backend/integration-tests/jmap/distributed-jmap-integration-tests/src/test/resources/mailetcontainer.xml
hungphan227: added IsOpenPaasAndDavSetup

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -0,0 +1,145 @@
...
+                            .flatMap(davCalendarObject -> updateDavCalendar(davCalendarObject, calendar, davUser))
+                            .switchIfEmpty(createDavCalendar(calendar, davUser)))
+                        .block());
+        } catch (Exception e) {
+            LOGGER.error("Error while handling calendar in mail {}", mail.getName(), e);
+        }
+    }
+
+    private Mono<Void> updateDavCalendar(DavCalendarObject davCalendarObject, Calendar calendar, DavUser davUser) {
+        // TODO Rewrite this method to update the calendar event
```
Arsnael, line 126: TODO not done spotted?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -0,0 +1,145 @@
...
+import com.linagora.tmail.dav.EventUid;
+
+import net.fortuna.ical4j.model.Calendar;
+import net.fortuna.ical4j.model.Property;
+import net.fortuna.ical4j.model.component.VEvent;
+import net.fortuna.ical4j.model.property.Uid;
+import reactor.core.publisher.Mono;
+
+public class CalDavCollect extends GenericMailet {
+    public static final class CalendarClients {
```
Arsnael, line 53: Is that really necessary? Can't we just inject directly davClient and davUserProvider? If missing cause not using OpenPaasModule it will throw directly at Tmail start anyways? I believe other mailets function like this..?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
hungphan227: To make it simple, I think we can simply do the following:
- Get userId (or openpassId)
- Build the URL
- send PUT to dav server

In case the METHOD is CANCEL, we will send DELETE to dav sever

There are other special cases which I am not sure if they could occur. For example, an organizer in gmail update the calendar event twice. However, the email of the second update is sent to linagora before the email of the first update. The caldavcollect mailet will mistakenly send PUT to dav server with first update after the first update. The user would see calendar event content of the first update instead of second update. Probably we can check this case in other ticket.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
vttranlina: The original ticket only covered CREATE and CANCEL events. 
@hungphan227 is likely want to UPDATE events (e.g., dtstart, dtend, location) in this pr, 
not sure it is necessary or not.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/dav/DavClient.java
```diff
@@ -66,11 +65,11 @@
 
 public class DavClient {
     public static final int MAX_CALENDAR_OBJECT_UPDATE_RETRIES = 5;
+    public static final String CALENDAR_PATH = "/calendars/";
```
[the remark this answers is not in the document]
vttranlina: this variable already exists in master code.
https://github.com/linagora/tmail-backend/blob/master/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/dav/DavClient.java#L73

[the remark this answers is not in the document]
Arsnael: And... is that a reason to not fix a code practice that we failed to see not being respected before?

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
Arsnael: If in a following PR, fair!

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
hungphan227: We could use LAST-MODIFIED field in ics file to check if the ics file is the latest (by comparing with ics stored in dav server) ---> We still need to getCalendarObject before sending PUT to dav server.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/dav/DavClient.java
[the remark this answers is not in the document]
hungphan227: > this variable already exists in master code. https://github.com/linagora/tmail-backend/blob/master/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/dav/DavClient.java#L73

I need to access it from other class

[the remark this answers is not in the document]
hungphan227: > Actually technically public static final variables should be under the private static final ones :)

In our project, at some places it is above, some places it is under. Anyway will update.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
vttranlina: The code seems to create a lazy bean.
It will look more pretty:

image

chibenwa (changes requested):
This PR ships dangerous changes that would utterly break the behaviour of event collection. Please let's think about 'the big picture' and what the code does.

I also expect from the team better than just cosmetic comments but I expect people to be able to catch structural defects.

Finally I think the team shall be doing a pair or even a MOB on the topic.

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/CalDavCollectIntegrationTest.java
hungphan227: This test just send email then process it --> trigger caldavcollect. It has nothing to do with blobresolver. The fake stuff here is just for required binding.

tmail-backend/integration-tests/jmap/distributed-jmap-integration-tests/src/test/resources/mailetcontainer.xml
hungphan227: Removed IsOpenPaasAndDavSetup. If we do not use openpass, we will not use this mailet as well

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -0,0 +1,132 @@
...
+            mail.getRecipients()
+                .forEach(mailAddress -> davUserProvider.provide(Username.of(mailAddress.asString()))
+                    .flatMap(davUser -> {
+                        EventUid eventUid = getEventUid(calendar);
+                        if (Method.VALUE_CANCEL.equals(calendar.getMethod().getValue())) {
+                            return deleteDavCalendar(davUser, eventUid);
+                        } else {
+                            return davClient.getCalendarObject(davUser, eventUid)
+                                .flatMap(davCalendarObject -> updateDavCalendar(davCalendarObject, calendar, davUser))
+                                .switchIfEmpty(createDavCalendar(calendar, davUser));
```
vttranlina, line 99: Please recheck it
IDE notice:
`Calling transform method switchIfEmpty on Mono<Void> will not have any effect `

quantranhong1999:
https://james-jenkins.lin-saas.com/blue/organizations/jenkins/Tmail%20build/detail/PR-1618/5/tests

Seems related.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/dav/DavClient.java
```diff
@@ -174,11 +173,10 @@ private Mono<Void> doUpdateCalendarObject(String username, DavCalendarObject upd
             .responseSingle((response, responseContent) -> handleCalendarObjectUpdateResponse(updatedCalendarObject, response));
     }
 
-    @VisibleForTesting
     public Mono<Void> createCalendar(String username, URI uri, Calendar calendarData) {
         return client.headers(headers -> headers.add(HttpHeaderNames.CONTENT_TYPE, "text/plain")
                 .add(HttpHeaderNames.AUTHORIZATION, authenticationToken(username)))
-            .request(HttpMethod.PUT)
+            .request(HttpMethod.valueOf("MKCALENDAR"))
```
vttranlina, line 179: Can you add basic tests for `createCalendar` ` deleteCalendar` method 
in the `CalDavEventRepositoryTest` ?
It is using "real" openpaas extension (not mock)

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
hungphan227: switchIfEmpty is for davClient.getCalendarObject

quantranhong1999: `updateDavCalendar` could return `Mono<Void>`...

hungphan227: > `updateDavCalendar` could return `Mono<Void>`...

so what?

vttranlina: I rechecked again, it looks like bug of IDE, it warning incorrect

vttranlina (approved):

hungphan227:
Ticket to update event https://github.com/linagora/tmail-backend/issues/1649

Arsnael (approved):

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/OpenPaasAmqpForwardAttribute.java
```diff
@@ -68,6 +68,8 @@
  * <p>
  * @see OpenPaasAmqpForwardAttributeConfig
  */
+
+@Deprecated
```
chibenwa, line 72: Please put in a comment why it is deprecated and when removal is expected (after 1.0.5)

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/IsOpenPaasRabbitMqSetup.java
```diff
@@ -28,6 +28,7 @@
 import org.apache.mailet.Mail;
 import org.apache.mailet.base.GenericMatcher;
 
+@Deprecated
```
chibenwa, line 31: Remove is straight away it is used only in tests

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -0,0 +1,132 @@
...
+                        }
```
chibenwa, line 100: Extract this in another method.

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -0,0 +1,132 @@
...
+                    }).block());
+        } catch (Exception e) {
+            LOGGER.error("Error while handling calendar in mail {}", mail.getName(), e);
+        }
+    }
```
chibenwa, line 105: Error handling sucks

If I send an email with ICS to bob and alice and that somehow we fail to update bob calendar, there is no reasons not to update Alice calendar

Please adapt exception handling accordingly.

chibenwa (changes requested):
Still missed documentation

In https://github.com/linagora/tmail-backend/blob/master/docs/modules/ROOT/pages/tmail-backend/configure/openpaas.adoc

tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/OpenPaasAmqpForwardAttribute.java
hungphan227: A second thought: should we mark this as deprecated after https://github.com/linagora/tmail-backend/issues/1649 is done? Caldavcollect mailet may not work well without event update capability.

chibenwa: Please keep it deprecated

Deprecated do not mean removed and we will work on #1649 ASAP

hungphan227:
> Another point of attention is handling recurrence
> 
> Bob invites ALICE to a meeting every MONDAY 7am BOB modifies the meeting on 7/04 to 6am BOB modifies the meeting on 14/04 to 8am
> 
> We shall be able to get the correct events in the calendar.
> 
> I'm creating a test event, before approuving I want to be sure that is is correctly handled.

This case will be handled in update event ticket https://github.com/linagora/tmail-backend/issues/1649. Any update to recurring event will be added to original ics file in sabre dav (tested this fact with https://github.com/linagora/tmail-backend/blob/master/tmail-backend/tmail-third-party/openpaas/docker-compose-carddav-sample.yml)

chibenwa (approved):

vttranlina (approved):

PR #1653, opened 2025-03-26
Twake Calendar Side Service: People search

chibenwa:
Bootstraps people API on top of opensearch


chibenwa:
Screenshot from 2025-03-26 23-30-41

PR #1657, opened 2025-03-28
WIP ISSUE-1585 Fix connections in RedisEventBusClientFactory

Arsnael:
Master replica and sentinel seem to use a redis standalone client setup for some reason, I think could be a reason for sentinel master failover issue? 

Just a draft for now, as core refactoring is underway which i hope should change and ease integration => https://github.com/apache/james-project/pull/2690

### DIFF
--- a/tmail-backend/event-bus-redis/src/main/scala/org/apache/james/events/RedisEventBusClientFactory.scala
+++ b/tmail-backend/event-bus-redis/src/main/scala/org/apache/james/events/RedisEventBusClientFactory.scala
@@ -30,13 +30,9 @@ class RedisEventBusClientFactory @Singleton() @Inject()
 (redisConfiguration: RedisConfiguration, redisClientFactory: RedisClientFactory) {
   val rawRedisClient: AbstractRedisClient = redisConfiguration match {
     case standaloneConfiguration: StandaloneRedisConfiguration => redisClientFactory.createStandaloneClient(standaloneConfiguration)
-    case masterReplicaRedisConfiguration: MasterReplicaRedisConfiguration =>
-      redisClientFactory.createStandaloneClient(new StandaloneRedisConfiguration(masterReplicaRedisConfiguration.redisURI.value.last,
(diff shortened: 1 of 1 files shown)


tmail-backend/event-bus-redis/src/main/scala/org/apache/james/events/RedisEventBusClientFactory.scala
```diff
@@ -30,13 +30,9 @@ class RedisEventBusClientFactory @Singleton() @Inject()
...
-    case masterReplicaRedisConfiguration: MasterReplicaRedisConfiguration =>
-      redisClientFactory.createStandaloneClient(new StandaloneRedisConfiguration(masterReplicaRedisConfiguration.redisURI.value.last,
-        masterReplicaRedisConfiguration.useSSL, masterReplicaRedisConfiguration.mayBeSSLConfiguration, masterReplicaRedisConfiguration.ioThreads, masterReplicaRedisConfiguration.workerThreads))
+    case masterReplicaRedisConfiguration: MasterReplicaRedisConfiguration => redisClientFactory.createMasterReplicaClient(masterReplicaRedisConfiguration)
     case clusterRedisConfiguration: ClusterRedisConfiguration => redisClientFactory.createClusterClient(clusterRedisConfiguration)
-    case sentinelRedisConfiguration: SentinelRedisConfiguration =>
-      redisClientFactory.createStandaloneClient(new StandaloneRedisConfiguration(sentinelRedisConfiguration.redisURI,
-        sentinelRedisConfiguration.useSSL, sentinelRedisConfiguration.mayBeSSLConfiguration, sentinelRedisConfiguration.ioThreads, sentinelRedisConfiguration.workerThreads))
+    case sentinelRedisConfiguration: SentinelRedisConfiguration => redisClientFactory.createSentinelClient(sentinelRedisConfiguration)
   }
```
chibenwa, line 36: IMO it would make more sense to @Inject the AbstractRedisClient and not the RediClientFactory and the conf to know the actual topology:

 - Less code duplication
 - We would only use a single driver event loop and not one per use case.

Arsnael: Still requires refactoring on james then, but that goes with your proposition on James side PR. I will take a look at it.

PR #1706, opened 2025-04-15
Tests for Redis event bus when Redis standalone restart

quantranhong1999:
cf https://github.com/linagora/james-project-private/issues/1052

## Summary

**TL;DR:** Basically, Redis event bus continues to function well after a Redis standalone restart.

- The **Lettuce driver** [handles reconnection automatically](https://javadoc.io/static/io.lettuce/lettuce-core/6.5.5.RELEASE/io/lettuce/core/protocol/ConnectionWatchdog.html) when Redis becomes available again after a restart, so there is **no need to restart TMail**.

### Impact of Redis Restart on the Event Bus

Overall, the impact is minimal:

#### 🔹 Users without active notification subscriptions before the Redis restart:
- No issues.
- They can subscribe to notifications normally after the restart, and receive notifications well.

#### 🔹 Users with active notification subscriptions before the Redis restart:
- Subscriptions are lost after the Redis restart.
- Users will need to **re-subscribe**, either via **IMAP SELECT** or refreshing the **WebSocket push** connection for example.
- The previous registrations are **automatically cleaned up** when the connection is closed by TMail, allowing new subscriptions to be properly stored in Redis (using `SADD` on a Redis Set).

#### ⚠️ Rare Edge Case:
- If cleanup does **not** occur as expected, even new subscriptions might fail.
- This happens because Redis eventbus ignores duplicate registration keys ([relevant code](https://github.com/linagora/tmail-backend/blob/ec13090ac17601bd13651ee87508f14c412a68ee/tmail-backend/event-bus-redis/src/main/java/org/apache/james/events/RedisKeyRegistrationHandler.java#L149)).

     ✅ Potential solutions for this edge case (if needed):
    - Enable **Redis AOF** to persist registration keys.
    - Always rebind the registration key (adds an extra `SADD` per subscription).
    
    However, since this is a very rare scenario, we can just skip it IMO.

### DIFF
--- a/tmail-backend/event-bus-redis/src/test/java/org/apache/james/events/RedisEventBusServiceTest.java
+++ b/tmail-backend/event-bus-redis/src/test/java/org/apache/james/events/RedisEventBusServiceTest.java
@@ -20,19 +20,26 @@ package org.apache.james.events;
 
 import static org.apache.james.events.EventBusTestFixture.EVENT;
 import static org.apache.james.events.EventBusTestFixture.KEY_1;
+import static org.apache.james.events.EventListener.ExecutionMode.ASYNCHRONOUS;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.awaitility.Awaitility.await;
+import static org.awaitility.Durations.TEN_SECONDS;
 
 import java.time.Duration;
 import java.util.concurrent.TimeoutException;
 
 import org.apache.james.backends.rabbitmq.RabbitMQExtension;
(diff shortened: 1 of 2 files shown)


chibenwa:
> However, since this is a very rare scenario, we can just skip it IMO.

Ok

> Which is not suitable to merge. Therefore, I only put this PR as a draft for review.

Those testa are imo valuable we can have a dedicate redis standalone fixed port extension and @Disable the test as part of the CI so that re-running it later is easy?

Arsnael:
Read it, nothing to add

quantranhong1999:
> Those testa are imo valuable we can have a dedicate redis standalone fixed port extension and @disable the test as part of the CI so that re-running it later is easy?

I added a comment on how to configure the fixed port and Disabled the tests.

chibenwa (approved):

PR #1724, opened 2025-04-24
Should not send request to caldav server when recipient is not part of the attendee list

Arsnael:
Draft for now, as the fixup! with the extra test allowed me to catch an other issue.

In fact ICALToJsonAttribute is storing json attributes per each recipient...

In CalDavCollect we process each json attribute... and loop on each mail recipient. 

So on that extra test I still can see a 400 call to caldav but for the correct receiver user because... we do the request twice in that scenario. First request is fine, second is a fail

We should just likely only loop on json attributes and extract the recipient from it, and not looping again on the list of recipients.

Will fix tomorrow.

Don't hesitate to give early feedback.

I thought as well about filtering on ICALToJsonAttribute but that implies a james code change, which I'm not sure is justified.

### DIFF
--- a/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
+++ b/tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
@@ -83,21 +83,23 @@ public class CalDavCollect extends GenericMailet {
     }
 
     private void handleCalendarInMail(byte[] json, Mail mail) {
-        String icalContent = extractIcalContent(json, mail.getName());
+        JsonNode jsonNode = convertToJson(json, mail.getName());
 
-        if (!icalContent.isEmpty()) {
-            mail.getRecipients()
-                .forEach(mailAddress -> {
-                    try {
-                        if (isRecipientAttendee(mailAddress, icalContent)) {
-                            davUserProvider.provide(Username.of(mailAddress.asString()))
-                                .flatMap(davUser -> synchronizeWithDavServer(json, davUser))
-                                .block();
-                        }
-                    } catch (Exception e) {
-                        LOGGER.error("Error while handling calendar in mail {} with recipient {}", mail.getName(), mailAddress.asString(), e);
-                    }
-                });
+        String icalContent = jsonNode.path("ical").asText();
+        String recipient = jsonNode.path("recipient").asText();
+
+        if (!icalContent.isEmpty() && !recipient.isEmpty()) {
+            try {
+                MailAddress mailAddress = new MailAddress(recipient);
+
+                if (isRecipientAttendee(mailAddress, icalContent)) {
+                    davUserProvider.provide(Username.of(mailAddress.asString()))
+                        .flatMap(davUser -> synchronizeWithDavServer(json, davUser))
+                        .block();
+                }
+            } catch (Exception e) {
+                LOGGER.error("Error while handling calendar in mail {} with recipient {}", mail.getName(), recipient, e);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


tmail-backend/tmail-third-party/openpaas/src/main/java/com/linagora/tmail/mailet/CalDavCollect.java
```diff
@@ -79,21 +83,45 @@ public void service(Mail mail) throws MessagingException {
...
+            return rootNode.get("ical").asText();
+        } catch (Exception e) {
+            LOGGER.error("Error while handling calendar in mail {}", mailName, e);
+            return "";
+        }
+    }
+
+    private boolean isRecipientAttendee(MailAddress mailAddress, String icalContent) {
+        return icalContent.toLowerCase().contains(mailAddress.getLocalPart().toLowerCase());
+    }
```
chibenwa, line 126: We need to be smarter and handle calendar data as structured data using ICAL4J

Arsnael: I find that a bit overkill IMO as we already parsed and processed the calendar with Ical4j earlier in the pipeline, feels like a lot of extra processing to make it smarter and just check if recipient is part of the attendee list?

But I will give a shot

tmail-backend/tmail-third-party/openpaas/src/test/java/com/linagora/tmail/CalDavCollectIntegrationTest.java
```diff
@@ -291,6 +297,35 @@ void mailetShouldCallDavSeverToCreateNewCalendarObject(@TempDir File temporaryFo
...
+        String mail = generateMail("template/emailWithAliceInviteBob.eml.mustache", generateEmailTemplateData(sender, receiver, mimeMessageId, mimeMessageId));
+
+        sendMessage(sender, notInvited, mail, mimeMessageId);
+
+        DavCalendarObject result = davClient.getCalendarObject(new DavUser(notInvited.id(), notInvited.email()), new EventUid(mimeMessageId)).block();
+        assertThat(result).isNull();
+    }
+
+    @Test
+    void mailetShouldCallOrNotCallDavServerToCreateNewCalendarObjectIfRecipientIsOrNotInvited(@TempDir File temporaryFolder) throws Exception {
```
hungphan227, line 312: i don't understand the test name

Arsnael: You have user1, user2, user3

User1 sends a mail to user2 and user3 with an ics calendar invite. But only user2 is an attendee, user3 is not.

CalDav should be called for user2 but not user3.

More clear? I concede the name is a bit confusing but if you want to propose a better name :)

hungphan227: mailetShouldOnlyCallDavServerToCreateNewCalendarObjectForInvitedRecipientsWhenMailContainsBothInvitedRecipientsAndUninvitedRecipients

hungphan227: `softly.assertThat(result2).isNull();`

I think this does not mean much because even in case dav server is called, it is still null

Arsnael: > I think this does not mean much because even in case dav server is called, it is still null

Correct. You can see when running the test locally if you get the 400 or not though.

quantranhong1999 (commented):
Read it.

chibenwa (approved):

quantranhong1999 (approved):

Arsnael:
> Related or not?

Likely not first view, green locally for me as well. I think it is flaky

PR #1730, opened 2025-04-29
ISSUE-1727 JMAP DownloadAll: Parse message for email body parts and r…

Arsnael:
…esolve blobs

DownloadAllContract tests are green, but I ain't 100% sure that this is what was wanted. Don't hesitate to comment otherwise


quantranhong1999 (commented):
I just wonder if we should keep using `AttachmentManager` for faster attachment retrieval and fallback to parsing the full message if needed.

chibenwa:
> I just wonder if we should keep using AttachmentManager for faster attachment retrieval and fallback to parsing the full message if needed.

That looks like a very good idea if you pull it off

Arsnael:
I can try to have a shot at it sure

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/routes/DownloadAllRoutes.scala
```diff
@@ -156,24 +159,32 @@ class DownloadAllRoutes @Inject()(@Named(InjectionKeys.RFC_8621) val authenticat
     val id: String = request.param(emailIdParam)
     SMono.fromCallable(() => messageIdFactory.fromString(id: String))
       .onErrorResume(e => SMono.error(MessageNotFoundException(id, e)))
-      .flatMap(messageId => SFlux(messageIdManager.getMessagesReactive(ImmutableList.of(messageId), FetchGroup.HEADERS_WITH_ATTACHMENTS_METADATA, mailboxSession))
+      .flatMap(messageId => SFlux(messageIdManager.getMessagesReactive(ImmutableList.of(messageId), FetchGroup.FULL_CONTENT, mailboxSession))
```
chibenwa, line 162: Why we need to change this?

We can still iterate the LoadedAttachment (which are badly named as they are `MessageAttachmentMetadata`) and simply load their content using the AttachmentId and the BlobResolvers?

chibenwa (changes requested):
Unecessarily inneficient

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/routes/DownloadAllRoutes.scala
Arsnael: I thought the pitch was :  attachments in the S3 blobstore got deleted by mistake? What was deleted exactly then? Rather confused with the description of the ticket then

Arsnael:
> For safety we shall read the full email, parse it as EmailBodyPart then zip all the attachments together.

I think was more or less my first version... hopefully I saved it on a local backup branch. Let me see.

Arsnael:
There.

I will try to add asked tests (attachmentv2 truncated table... and inlined images)
And will see if I can do as @quantranhong1999 suggested at first, rely on attachment table first like now, and then fallback to full message read if unsuccessful.

Arsnael:
Added the test for distributed where we delete attachments in cassandra table and still are able to retrieve attachments

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/routes/DownloadAllRoutes.scala
```diff
@@ -156,24 +159,32 @@ class DownloadAllRoutes @Inject()(@Named(InjectionKeys.RFC_8621) val authenticat
...
-  private def getBlob(attachmentId: AttachmentId, mailboxSession: MailboxSession): SMono[Blob] =
-    SMono(attachmentManager.getAttachmentReactive(attachmentId, mailboxSession))
-      .flatMap(attachmentMetadata => SMono(attachmentManager.loadReactive(attachmentMetadata, mailboxSession))
-        .map(content => AttachmentBlob(attachmentMetadata, content)))
+  private def getAttachments(messageResult: MessageResult): SFlux[ParsedAttachment] =
+    SFlux.fromIterable(messageParser.retrieveAttachments(messageResult.getFullContent.getInputStream)
+      .getAttachments
+      .asScala)
+
+  private def getBlob(attachment: ParsedAttachment, messageId: MessageId): SMono[Blob] = // blobResolver?
```
quantranhong1999, line 181: remove comment?

quantranhong1999 (commented):
Read it.

> Also please add a test for inlined images: they shall not be included in downloadAll feature.

Still missing as I see.

Arsnael:
> Still missing as I see.

Yes I know still working on things :)

tmail-backend/integration-tests/jmap/distributed-jmap-integration-tests/src/test/java/com/linagora/tmail/james/DistributedDownloadAllRouteTest.java
```diff
@@ -64,11 +84,44 @@ public class DistributedDownloadAllRouteTest implements DownloadAllContract {
...
+            .overrideWith(new CassandraAttachmentProbeModule()))
         .build();
 
     @Override
     public MessageId randomMessageId() {
         return MESSAGE_ID_FACTORY.of(Uuids.timeBased());
     }
+
+    @Test
+    void downloadAllShouldWorkAfterAttachmentMetadataGotDeletedByMistake(GuiceJamesServer server) throws MailboxException {
```
chibenwa, line 97: (Not really by mistake it was intentionnal!)

:-)

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/routes/DownloadAllRoutes.scala
chibenwa: Yep no direct access is 100% better than blob resolver.

chibenwa, line 181:
```suggestion
  private def getBlob(attachment: ParsedAttachment, messageId: MessageId): SMono[Blob] = 
```

Arsnael: Sorry I forced push as I did fix it in a commit and was lazy to rebase :)

Arsnael:
Alright I added:
- well initial way of packaging all downloads and falling back to full read if empty (let me know if ok, but it works)
- indeed inlined attachments were part of download all, added tests and proper fix

Ready for final review!

chibenwa (approved):

PR #1739, opened 2025-05-07
ISSUE-1732 Increase Mailbox/clear deletion batch

quantranhong1999:
20k mails, before: 15 seconds
20k mails, after: 10 seconds


Arsnael (approved):

chibenwa (approved):
Marginal gains but still good to take.

PR #1741, opened 2025-05-09
OIDC token cache

Arsnael:


chibenwa:
(Will review when base PR will get merged.)

chibenwa (approved):
Arf easy, no redis stuff yet :-)

Good with me.

Arsnael:
Rebased

quantranhong1999 (approved):
Otherwise LGTM.

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/oidc/OidcTokenCache.java
```diff
@@ -0,0 +1,27 @@
...
+ *  warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR         *
+ *  PURPOSE. See the GNU Affero General Public License for          *
+ *  more details.                                                   *
+ ********************************************************************/
+
+package com.linagora.tmail.james.jmap.oidc;
+
+import reactor.core.publisher.Mono;
+
+public interface OidcTokenCache {
```
quantranhong1999, line 23: IMO these classes should be put in the `jmap-extensions-api` module instead of the `jmap-extensions` module.

Arsnael: would require further refactoring/moves as `jmap-extensions` relies on `jmap-extensions-api` and not the other way around. Should we?

quantranhong1999: We can let that later if needed IMO. Let's go fast first.

quantranhong1999 (approved):

PR #1790, opened 2025-06-10
Updating demo to use the new extensionMecanism

### LINKED ISSUE #1789: Fix demo
## Why?

Extensions updates and new worrk on AI did break tmail backend demo.

## What ?

Update the demo app accordingly

## Definition of done

Able to start the demo with latest memory image and the IA extension

AlaeMghirbi:
closes #1789 
adding aibot config to ai.properties and removing it from mailetContainer.xml, adding extension modules for aiBot
updating the jar with the new aibot with dependencies

### DIFF
--- a/demo/tmail/mailetcontainer.xml
+++ b/demo/tmail/mailetcontainer.xml
@@ -96,13 +96,8 @@
             <mailet match="All" class="com.linagora.tmail.mailets.TmailLocalDelivery">
                 <consume>false</consume>
             </mailet>
+            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.AIBotMailet"/>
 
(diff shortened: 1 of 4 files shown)


demo/tmail/mailetcontainer.xml
```diff
@@ -96,13 +96,6 @@
             <mailet match="All" class="com.linagora.tmail.mailets.TmailLocalDelivery">
                 <consume>false</consume>
             </mailet>
-
-            <!-- Put the OpenAIMailet after LocalDelivery so the GPT reply would come after the asking question -->
-            <mailet match="com.linagora.tmail.mailet.RecipientsContain=gpt@tmail.com" class="com.linagora.tmail.mailet.AIBotMailet">
```
chibenwa, line 101: Let's keep the mailet, no?

chibenwa: Yes but we need to keep the mailet, no?

PR #1826, opened 2025-07-10
[ENHANCEMENT] Load Keystore stuff as an extension

chibenwa:
This is not core-product and loading it as an external extension allow eg to not expose the JMAP capability and to not load the associated Cassandra data definition.


quantranhong1999:
Green build

Arsnael (approved):

PR #1834, opened 2025-07-16
ISSUE-1831 Refactor cassandra user table

### LINKED ISSUE #1831: Refactoring the notion of user in Twake mail
## Why?

In SaaS we may weant do add multiple  directions of user bound limits that we do not know to store well today.

We lack a notion of "Account" that would allow to mutualize knowlege about a specific user.

We have a collection of "tables" that revolve around the users:

 - `users`
 - `rate-limiting-plam`
 - `settings`
 
We may add other things like:
 - maximum size of a mail the user is allowed to send
 - maximum count of filters
 - maximum count of folders
 - is IMAP/SMTP active or not
 - ...

And lack a proper data structure for it.

Adding new Cassandra tables is a not good practice CF https://thelastpickle.com/blog/2020/11/25/impacts-of-many-tables-on-cassandra.html

## Proposition

Stop using James `users` data definition

Write our own `users` data definition (keep the same name to avoid having to migrate delegations...)

Add to it the coloumns we need. Migration plan: create the missing columns.

```
// Keep james Data structure
PK name TEXT
C realname TEXT 
C passwd TEXT // not used
C algorithm TEXT // not used
C authorized_users SET<TEXT>
C delegated_users SET<TEXT>
// merge with settings table - migration plan none as it is not used
C settings MAP<TEXT, TEXT>
C settings_state UUID
// merge with rate limiting - migration plan none as it is not used
C plan_id TEXT
```

### How

 - Drop the data definitions for the user table in James
 - Create tmail's `users` table data definition in `distributed` app and bind it
 - Duplicate it in each and every call point - existing DAOs would blindly use the new table and stop using their old table. 
 - Make it work.

And do something similar for PG.

quantranhong1999:
resolve #1831


Arsnael:
> From the ticket description: migration plan none as it is not used.

Missed that point in the ticket sorry, good to me then!

Issue #1840, opened 2025-07-21
Back the "get more" link in the SaaS offer


chibenwa:
## Why?

As part of our SaaS offer we want to display a `get more` link redirecting to the pay-wall for free user to extend their plan and pay more.

This needs to work on SaaS and mobile.

We shall avoid displaying this info to users that paid.

## How?

Use a JMAP capability to advertize the 'get more' link and message within the web/mobile app. Eg:

```
"com:linagora:params:saas": {
  "payingUser": false
}
```

Based on this  the mobile app will hard code URL and message if it needs to be displayed to the user in order to advertize the paywall.

Use a column of the `users` table in order to write a `PayingUserPredicate` interface:

```
interface PayingUserPredicate {
    Publisher<Boolean> isPaying(Username username);

   Publisher<Void> setPaying(Username username, boolean value); // to be called by the rabbit listener
}
```

Implement it for PG and Cassandra and memory.

Based on it write a SaaSCapability for JMAP.

Define a guice module, not included by default in each tmail App, in order to define injects for `SaaSCapability` and its dependency (we leverage extensions.properties for loading it or not.

We will then add a helm chart capability to define extra modules in extensions.properties.

chibenwa:
# Step 1

```
interface PayingUserPredicate {
    Publisher<Boolean> isPaying(Username username);
}
``` 

And a `NaivePayingUserPredicate` that always return true.

Implement `com:linagora:params:saas` based on `PayingUserPredicate`

Write a `NaiveSaaSModule` being NaivePayingUserPredicate + add support for com:linagora:params:saas

We will then add a helm chart capability to define extra modules in extensions.properties.

# Step 2

Implement PayingUserPredicate for PG and Cassandra and memory.

Define a guice module, not included by default in each tmail App, in order to define injects for SaaSCapability and its dependency (we leverage extensions.properties for loading it or not.

chibenwa:
> However, a detail, the button should probably not be displayed based on whether the service is free or paid, but rather based on the user's offer. If the user can increase their space, even if it is a paid service, the button will be displayed.

Seems fair.

So we could store a `plan` property being a constant with value `free`, `standard`, (possibly other values), `premium` and decide to not display the "getMore link" if and only the user is in premium plan?

chibenwa:
> Are you ok with Cassandra, Postgres, and memory implementations in the same saas Maven extension module, to ease injection with just 1 jar?

No.

[...]

PR #1861, opened 2025-08-04
ISSUE-1840 Cassandra implementation for SaaS capability

quantranhong1999:
part of #1840


Arsnael (approved):
Good approach with avoiding creating unnecessary columns for non saas deployments!

Issue #1888, opened 2025-09-04
SaaS rate limiting: Handle rate limiting definition as a subproperty of the users table set by the SaaS connector


quantranhong1999:
```
{
"userEmail": "btellier@twake.app",
 ...
"mail": {
  ...
  "mailsSentPerMinute": 100,
  "mailsSentPerHours": 1000,
  "mailsSentPerDays": 2000,
  "mailsReceivedPerMinute": 100,
  "mailsReceivedPerHours": 1000,
  "mailsReceivedPerDays": 2000,
 }
}
```

The rate limits could be stored as a JSON column in the database, as part of `SaaSAccountRepository`.

  TODO:
  - implement storage layer: memory, Cassandra, PostgreSQL
  - document `upgrade-instructions.md`
  - implement SaaS connector part: deserialize rate limiting, store the rate limiting info

quantranhong1999:
rework: https://github.com/linagora/tmail-backend/pull/1900

PR #1895, opened 2025-09-05
ISSUE-1888 Implement SaaS rate limiting storage

quantranhong1999:
resolve https://github.com/linagora/tmail-backend/issues/1888


quantranhong1999:
see https://github.com/linagora/tmail-backend/pull/1900 instead

PR #1902, opened 2025-09-09
ISSUE-1888 SaaSSubscriptionConsumer should persist rate limiting

quantranhong1999:
resolve #1888 
Implement the SaaS connector part + username change.
For user data deletion, it would be implicitly a part of user record deletion.

PR #1917, opened 2025-09-15
ISSUE-1887 Implement username change step for SaaS account

### LINKED ISSUE #1887: [SaaS] SaaSAccountRepository: username change and user data deletion
## DoD
- [ ] Username change and user data deletion are supported for SaaS data.

quantranhong1999:
resolve #1887

### DIFF
--- a/docs/modules/ROOT/pages/tmail-backend/webadmin.adoc
+++ b/docs/modules/ROOT/pages/tmail-backend/webadmin.adoc
@@ -332,10 +332,10 @@ Implemented extra steps for TMail are:
 
 - `ContactUsernameChangeTaskStep`: migrates contacts from old user to new user
 - `PGPKeysUsernameChangeTaskStep`: migrates PGP public keys from old user to new user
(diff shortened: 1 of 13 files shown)


Arsnael (approved):

Arsnael:
> Please confirm we need something for rate limiting too.

I would think so too. I would be ok if it's on an other PR though as well

quantranhong1999:
> Please confirm we need something for rate limiting too.
> I would think so too. I would be ok if it's on an other PR though as well

Hi, we did implement that as part of https://github.com/linagora/tmail-backend/pull/1902/commits/6262b8e7b60212b4bdbe02013f7adcea5b68b714

Arsnael:
> Hi, we did implement that as part of https://github.com/linagora/tmail-backend/commit/6262b8e7b60212b4bdbe02013f7adcea5b68b714

I think it does the username change. What about the delete saas account?

quantranhong1999:
> I think it does the username change. What about the delete saas account?

It would be a part of user (row) deletion.
cf https://github.com/linagora/tmail-backend/issues/1887#issuecomment-3248437667

PR #1962, opened 2025-10-13
ISSUE-1924 Saas domain creation

Arsnael:
Still missing the health check


chibenwa (commented):
Very clean history, on an overall very complete changeset.

Thanks for putting this together.

I think there might be 2-3 little ommissions:

 - [x] Do we have a reconnection handler for this queue ?
 - [x] Are we monitoring the dead-letter queue for emptiness ? (Find a way to patch SaaSSubscriptionDeadLetterQueueHealthCheckto monitor 2 queues?)
 - [x] Are we monitoring the queue for consumers? (bundle SaaSSubscriptionQueueConsumerHealthCheck to monitor 2 queues?)

Appart from this nothing to say.

Issue #1927, opened 2025-09-17
[SaaS] Domain based rate limiting Webadmin management


chibenwa:
Webadmin endpoints for https://github.com/linagora/tmail-backend/issues/1926

Why: allow reuse the rate limiting component for on prem

```
curl -XPUT http://ip:port/domains/{domain}/ratelimits
 -H "Content-Type: application/json"
 -d '{
    "mailsSentPerMinute": 10,
    "mailsSentPerHours": 100,
    "mailsSentPerDays": 1000,
    "mailsReceivedPerMinute": 20,
    "mailsReceivedPerHours": 200,
    "mailsReceivedPerDays": 2000
}'
```

And 

```
curl -XGET http://ip:port/users/{username}/ratelimits

Returns the limits of a user:

{
    "mailsSentPerMinute": 10,
    "mailsSentPerHours": 100,
    "mailsSentPerDays": 1000,
    "mailsReceivedPerMinute": 20,
    "mailsReceivedPerHours": 200,
    "mailsReceivedPerDays": 2000
}
```

# Definition of done

 - [ ] Unit tests
 - [ ] Guice bindings
 - [ ] doc

Issue #1926, opened 2025-09-17
[SaaS] Domain based rate limiting


chibenwa:
# Why?

The admin panel will push to Twake mail domain scopped limits

# Spec

https://www.notion.so/linagora/ADR-021-B2B-RabbitMQ-communications-27062718bad1800fb8efe8e2e3098f10

```
{
  "domain": "linagora.com", 
  [...]
  "features":
    "mail": {
      [...]
      "mailsSentPerMinute": 100,
      "mailsSentPerHour": 1000,
      "mailsSentPerDay": 2000,
      "mailsReceivedPerMinute": -1, // -1 means ulimited, positive integer
      "mailsReceivedPerHour": -1,
      "mailsReceivedPerDay": -1,
    }
  }
}
```

`-1` meaning unlimited

Published on the `billing` exchange with the `domain.subscription.changed` routing key

 - [ ] Create a DAO for definition rate limits on domain. We can likely mutualise code a lot with the same DAO backing the `users` table (only change table name?)
 - [ ] Plug it into the Saas rabbit consumer
 - [ ] If a user has no rate limits fallback to the limit defined by the domain with Rate limiting mailets.

quantranhong1999:
> Create a DAO for definition rate limits on domain. We can likely mutualise code a lot with the same DAO backing the users table (only change table name?)

Likely, we can store the domain's rate limits by optionally adding rate limiting columns to the `domains` table. And
```java
public interface RateLimitingRepository {
    Publisher<Void> setRateLimiting(Username username, RateLimitingDefinition rateLimiting);

    Publisher<Void> setRateLimiting(Domain domain, RateLimitingDefinition rateLimiting);

    ...
}
```

chibenwa:
Yes that was exactly the idea

quantranhong1999:
Hi @Arsnael, I pushed the rate limit APIs for domain at https://github.com/linagora/tmail-backend/pull/1965 so it wont block this work.

Arsnael:
Cheers @quantranhong1999 :)

Arsnael:
https://github.com/linagora/tmail-backend/pull/1966

PR #1965, opened 2025-10-15
ISSUE-1927 Implement domain rate limit APIs

quantranhong1999:
part of #1927


quantranhong1999:
> I do not see this new DAO plugged onto the SaaS consumer, I presume this is for a follow up work.

Yes. I push the APIs early as part of https://github.com/linagora/tmail-backend/issues/1927, so it won't block https://github.com/linagora/tmail-backend/issues/1926.

Issue #1946, opened 2025-10-03
IdentityProvisionnerListener


chibenwa:
## Why?

Configuring LSC for identity is a pain.

We always forget. 

Let's explore alternatives!

## What ?

Write a `IdentityProvisionnerListener` listener that upon mailbox creation

Lookup directly the user into the LDAP to catch display name

And create the identity accordingly.

## Definition of done

With a LDAP no LSC is needed for  identity provisionning...

PR for helm chart

chibenwa:
> How about the domain contacts sync (in the demo too)?

Fair.

I removed this from the DOD.

chibenwa:
(We could imagine also have a DomaainContactProvisionner that provision contacts based on INBOX creation / deletion)

chibenwa:
Also PRed tmail-backend HELM chart as part of this work? (conditionally enabled, defaulting to true)

quantranhong1999:
helm PR: https://ci.linagora.com/linagora/lrs/saas/tools/helm-charts/tmail-backend/-/merge_requests/195

chibenwa:
Small comment on the helm PR: do we have any reason not to set this as the new default?

quantranhong1999:
> Small comment on the helm PR: do we have any reason not to set this as the new default?

Let's turn it on by default.

PR #1972, opened 2025-10-17
ISSUE-1954 Implement a listener for domain contact provisioning

Arsnael:


chibenwa (approved):
Nice to also have included deletion.

tmail-backend/jmap/extensions-api/src/main/java/com/linagora/tmail/james/jmap/contact/DomainContactProvisionerListener.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+        return Mono.empty();
+    }
+
+    private Mono<Void> addDomainContact(Domain domain, Username username) {
+        if (ignoredDomains.contains(domain)) {
+            return Mono.empty();
+        }
+        return Mono.fromCallable(() -> new ContactFields(username.asMailAddress(), "", ""))
+            .flatMap(contactFields -> Mono.from(contactSearchEngine.index(domain, contactFields)))
```
quantranhong1999: I propose to do better: lookup the firstname and surname for the user in LDAP.

That is what I am doing in https://github.com/linagora/tmail-backend/issues/1946

quantranhong1999: > the final decision on the ticket?

In the ticket description: `Similar than https://github.com/linagora/tmail-backend/issues/1946 but for domain contacts`

The other ticket used LDAP to get information. I was surprised to see that this does not need to use LDAP TBH.

PR #1973, opened 2025-10-17
ISSUE-1946 Implement IdentityProvisionListener

quantranhong1999:


chibenwa (commented):
Good work

tmail-backend/jmap/extensions/src/main/java/com/linagora/tmail/james/jmap/event/IdentityProvisionListener.java
```diff
@@ -0,0 +1,221 @@
...
+                    .doOnSuccess(identity -> LOGGER.info("Created default identity for user {} with display name {}", username.asString(), identity.name()))))
+            .then()
+            .onErrorResume(e -> {
+                LOGGER.error("Unexpected error during JMAP identity provisioning for user {}", event.getUsername(), e);
+                return Mono.empty();
+            })
+            .subscribeOn(ReactorUtils.BLOCKING_CALL_WRAPPER);
+    }
+
+    private SearchResultEntry searchLdap(Event event, Username username) throws LDAPSearchException {
```
Arsnael: Can drop event here? The only value you access from the event is username, that you pass as an argument already and use

Arsnael (approved):

PR #1979, opened 2025-10-22
ISSUE-1950 Implement FolderFilteringAction/get method

### LINKED ISSUE #1950: Implement FolderFilteringAction/get method
## Why

We would implement a JMAP `FolderFilteringAction/get` extension that returns the information about the filtering task.
```
[
    "FolderFilteringAction/get",
    {
        "ids": ["defghi"] // NOT null
    },
    "#0
]
```
Would return the id of the filtering task.

e.g.:
```
[
    "FolderFilteringAction/get",
    {
        "list": [
            {
                "id": "defghi",
                "status": "completed",
                "mailboxId": "abc123",
                "processedMessageCount": 9,
                "successfulActions": 9,
                "failedActions": 3,
                "maximumAppliedActionReached": false
            }
        ],
        "notFound": []
    },
    "#0"
]
```

## How

- Implement `FolderFilteringAction/get` method (we can take inspiration from `EmailRecoveryAction/get`)
- Contract tests

Note:
- method does not require accountId (aligned with `EmailRecoveryAction`)
- requires the Filter capability: `com:linagora:params:jmap:filter`
- we must not return tasks of other users
- we must return only filtering tasks (e.g. not return the Deleted message vault task)

quantranhong1999:
resolve #1950

### DIFF
--- a/docs/modules/ROOT/pages/tmail-backend/jmap-extensions/jmapFilters.adoc
+++ b/docs/modules/ROOT/pages/tmail-backend/jmap-extensions/jmapFilters.adoc
@@ -347,7 +347,7 @@ This method returns the underlying folder filtering tasks.
 Standard `/get` semantics. The `ids` property cannot be null.
 
 This method supports the following properties of the `FolderFilteringAction` object:
-`id`, `mailboxId`, `status`, `processedMessageCount`, `successfulActions`, `failedActions`, `maximumAppliedActionReached`.
+`id`, `status`, `processedMessageCount`, `successfulActions`, `failedActions`, `maximumAppliedActionReached`.
 
 Servers MUST ensure only the authenticated user's filtering tasks are returned.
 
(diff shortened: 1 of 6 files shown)


docs/modules/ROOT/pages/tmail-backend/jmap-extensions/jmapFilters.adoc
```diff
@@ -479,7 +480,7 @@ Server returns:
     "list": [
       {
         "id": "flt-task-001",
-        "mailboxId": "mailboxId123",
+        "mailboxName": "mailbox123",
```
quantranhong1999: Reason: we can not resolve `mailboxId` from `mailboxName` if the mailbox was deleted.
Frontend can use `mailboxName` to resolve the mailbox and know if the mailbox was deleted if needed.

Otherwise, we can add `mailboxId` to the task's additional information. Thoughts?

quantranhong1999: Ok. I will drop the latest commit.

> Otherwise, we can add mailboxId to the task's additional information

We will follow up that in a later PR.

PR #1986, opened 2025-10-27
[Build] Update the build timeout to 4h

Arsnael:
Been observing multiple builds failing on the 3h timeout lately. We keep having more code and tests it seems.


chibenwa (approved):
:-(

PR #2002, opened 2025-11-06
ISSUE-1984 Implement user quota reporter routes

### LINKED ISSUE #1984: [CUSTOMER] List users with quota exception
## Why

As part of a contract we allow user to create quota exceptions for his users and need to bill him for it.

## How

Webadmin endpoint in twake mail that list users with specific quota

## What

```
GET /quota/users?hasSpecificQuota

[
   {"user": "btellier@linagora.com", "storageLimit": 23878235, "countLimit" null},
   {"user":"zbellot@linagora.com, "storageLimit": 23878235, "countLimit" null}
...
]
```

Lists users with specific quota.

```
GET /quota/users/count?hasSpecificQuota

36
```

returns the total count of users that has specific quota

```
GET /quota/users/sum?hasSpecificQuota

{"storageLimit": 23878235, "countLimit" null}
```

Returns the sum of allocated extra quota (for each user `max(0, userQuota - domainQuota)`

## How

Write an interface `UserQuotaReporter` in `tmail-backend/data/data-extra-api`

Write implementation for `Cassandra` and `Postgres` reading if needed directly the concerned tables.

Regarding memory I suggest to implement a minimalistic implem to pass the tests (abstract method in test to add records in UserQuotaReporter in a per-implem specific way?)

quantranhong1999:
resolve #1984

TODO:
- [ ] fix the quota root -> username resolution
- [ ] docs

## Summary by CodeRabbit

* **New Features**
  * WebAdmin: three user quota report endpoints — list users with specific quotas, count them, and return aggregated extra-quota sums; available for Cassandra, Memory and PostgreSQL backends.
* **Documentation**
  * Added "User quota reports" section with endpoint docs, examples and responses.
* **Tests**
  * New unit and integration tests covering routes, backends and aggregated quota behaviors.

### DIFF
--- a/tmail-backend/data/data-extra-api/src/test/java/com/linagora/tmail/mailbox/quota/UserQuotaReporterContract.java
+++ b/tmail-backend/data/data-extra-api/src/test/java/com/linagora/tmail/mailbox/quota/UserQuotaReporterContract.java
@@ -52,7 +52,7 @@ public interface UserQuotaReporterContract {
     Username ALICE = Username.of("alice@domain.tld");
     QuotaRoot ALICE_QUOTA_ROOT = QuotaRoot.quotaRoot(MailboxConstants.USER_NAMESPACE + SEPARATOR + ALICE.asString(), Optional.of(DOMAIN_1));
     Username ANDRE_AT_DOMAIN_2 = Username.of("andre@domain2.tld");
-    QuotaRoot ANDRE_AT_DOMAIN_2_QUOTA_ROOT = QuotaRoot.quotaRoot(MailboxConstants.USER_NAMESPACE + SEPARATOR + ANDRE_AT_DOMAIN_2.asString(), Optional.of(DOMAIN_1));
+    QuotaRoot ANDRE_AT_DOMAIN_2_QUOTA_ROOT = QuotaRoot.quotaRoot(MailboxConstants.USER_NAMESPACE + SEPARATOR + ANDRE_AT_DOMAIN_2.asString(), Optional.of(DOMAIN_2));
     Username LOCAL_PART_ONLY = Username.of("localpartOnly");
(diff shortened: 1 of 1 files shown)


tmail-backend/integration-tests/webadmin/webadmin-integration-tests-common/src/main/java/com/linagora/tmail/integration/UserQuotaReporterRoutesIntegrationContract.java
```diff
@@ -0,0 +1,127 @@
...
+
+        assertThatJson(response)
+            .isEqualTo("""
+                [
+                    {
+                        "user": "bob@domain.tld",
+                        "storageLimit": 1000,
+                        "countLimit": null
+                    }
+                ]
```
quantranhong1999: Failing with response:
```json
[
    {
        "user": "#private&bob@domain.tld",
        "storageLimit": 1000,
        "countLimit": null
    }
]
```

I will attempt to resolve the correct username from the `QuotaRoot` instead.

chibenwa: QuotaRootResolver::associatedUsername ???

tmail-backend/webadmin/webadmin-mailbox/src/test/java/com/linagora/tmail/webadmin/quota/UserQuotaReporterRoutesTest.java
```diff
@@ -0,0 +1,403 @@
...
+                .contentType(ContentType.JSON)
+                .extract()
+                .body().asString();
+
+            assertThatJson(response)
+                .isEqualTo("""
+                    {
+                        "storageLimit": 300,
+                        "countLimit": 301
+                    }
```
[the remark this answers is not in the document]
chibenwa: Yes but from a billing perspective this is not exactly what we would want IMO

I think we are better off doing a sum without the unlimited values and then count unlimited values separately...

PR #2012, opened 2025-11-17
Align bouncycastle version with James

quantranhong1999:
Align bouncycastle version with James: https://github.com/apache/james-project/blob/0cd58041113e45c8346e9889c74fd60594f59462/pom.xml#L681

Food for thought: having two `bouncycastle.version`: one in James and one on TMail is really error-prone. I did have a very quick look, but did not find a way to reference both of them to the same property yet.

## Summary by CodeRabbit

* **Chores**
  * Updated a cryptographic library dependency to the latest stable version.


chibenwa (approved):

chibenwa:
Let's retag when merged....

PR #2085, opened 2025-12-31
Remove OpenPaasAmqpForwardAttribute and related deprecated code

Arsnael:
## Summary by CodeRabbit

* **Chores**
  * Removed deprecated AMQP forwarding mailet, its config, and the ICAL→JSON mailet plus associated integration tests and ICS test resources.

* **Refactor**
  * Introduced a ConfigurationPathFactory to manage test server configuration and calendar enablement; replaced prior OpenpaasTestUtils usage.

* **Tests**
  * Consolidated/simplified calendar test startup APIs and removed legacy contract-style AMQP calendar reply tests.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/CalendarEventSupportCapabilityContract.scala
+++ b/tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/CalendarEventSupportCapabilityContract.scala
@@ -37,8 +37,7 @@ trait CalendarEventSupportCapabilityContract {
 
   def stopJmapServer(): Unit
 
-  private def setUpJmapServer(calDavSupport: Boolean): Unit = {
-    val server = startJmapServer(calDavSupport)
+  private def provisionTestResources(server: GuiceJamesServer): Unit = {
     server.getProbe(classOf[DataProbeImpl])
       .fluent()
       .addDomain(DOMAIN.asString())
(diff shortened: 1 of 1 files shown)


tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/src/test/java/com/linagora/tmail/james/openpaas/OpenpaasTestUtils.java
```diff
@@ -29,15 +29,18 @@
 
 public class OpenpaasTestUtils {
 
-    public static Configuration.ConfigurationPath setupConfigurationPath(File workingDir) {
+    public static Configuration.ConfigurationPath setupConfigurationPath(File workingDir, boolean calDavSupport) {
```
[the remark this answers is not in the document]
Arsnael: Wasn't sure was necessary to go that far for tests but fair

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/CalendarEventSupportCapabilityContract.scala
chibenwa: Proposal is to abstract this away by passing `GuiceJamesServer` as a method parameter.
This would uncorelate resource provisionning from instanciation and thus clarify a lot the responsibilities.

(Single responsibility pattern : we are starting the server AND creating some resources)

=>

```
private def provisionTestResources(server: GuiceJamesServer): Unit = { ... }
```

Callers would instead of `setUpJmapServer(true)` do `provisionTestResources(startJmapServerWithCalendarSupport())`...

BTW it might be enough to kill the "boolean parameter" pattern.

chibenwa: >  that's just being very picky for calendar tests IMO

True.

But I believe the team can take this example to actually become better craftmans and spend more time on small refactorings that on the long run leads to better code.

Let's make it a habbit!

quantranhong1999 (approved):

PR #2131, opened 2026-01-31
ISSUE-2130 Fix JMAP Email/query subject filter to support substring matching

guimard:
## Summary
- Add `WildcardQuery` on `subject.raw` (case-insensitive) as a `should` clause alongside existing tokenized/ngram queries in `TmailCriterionConverter.convertSubject()`
- This ensures substring matching works for subjects containing special characters (`.`, `-`, `[`, `]`) per RFC 8621
- Added 3 new integration tests covering `example.com`, `nas-backup`, and `[nas-backup.example.com]` substring searches
- Updated existing ngram heuristic test to reflect that substring matching now works via wildcard

## Test plan
- [ ] `TmailOpenSearchIntegrationTest` passes (new + existing tests)
- [ ] Verify `subjectSearchShouldMatchSubstringWithSpecialCharacters`
- [ ] Verify `subjectSearchShouldMatchSubstringWithHyphen`
- [ ] Verify `subjectSearchShouldMatchSubstringWithBrackets`
- [ ] Verify updated ngram heuristic test now expects results

Fixes #2130

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2132, opened 2026-01-31
ISSUE-2130 Alternative: Use substring_analyzer for subject and address substring matching

### LINKED ISSUE #2130: JMAP Email/query subject filter does not perform substring matching
# JMAP Email/query subject filter does not perform substring matching

## Summary

The `subject` filter in `Email/query` appears to tokenize the search string on special characters (`.`, `-`, `[`, `]`) rather than performing substring matching as specified in RFC 8621.

## Environment

- **Server**: Apache James
- **Protocol**: JMAP (RFC 8620 / RFC 8621)

## Expected Behavior (RFC 8621)

According to [RFC 8621 Section 4.4.1](https://www.rfc-editor.org/rfc/rfc8621.html#section-4.4.1):

> **subject**: `String` - Looks for the text in the subject of the Email.

The specification indicates that this should be a substring match, similar to the `from`, `to`, and `text` filters which are described as searching for the string within the respective fields.

## Actual Behavior

The `subject` filter appears to tokenize the search string on special characters and performs a token-based match (all tokens must be present) rather than a substring match.

### Test Case 1: Search with dots in subject filter

**Given** an email with subject: `[nas-backup.example.com] Backup completed successfully`

**Request**:
```json
{
  "using": ["urn:ietf:params:jmap:core", "urn:ietf:params:jmap:mail"],
  "methodCalls": [
    ["Email/query", {
      "accountId": "...",
      "filter": { "subject": "example.com" }
    }, "0"]
  ]
}
```

**Expected**: Email is returned (substring "example.com" exists in subject)

**Actual**: No results returned

### Test Case 2: Search with hyphen in subject filter

**Given** emails with subjects:
- `[nas-backup.example.com] Backup completed` (contains literal "nas-backup")
- `Meeting with John Doe - Backup strategy for NAS` (contains "backup" and "nas" separately)

**Request**:
```json
{
  "using": ["urn:ietf:params:jmap:core", "urn:ietf:params:jmap:mail"],
  "methodCalls": [
    ["Email/query", {
      "accountId": "...",
      "filter": { "subject": "nas-backup" }
    }, "0"]
  ]
}
```

**Expected**: Only the first email is returned (contains literal substring "nas-backup")

**Actual**: Both emails are returned because they contain "nas" AND "backup" as separate tokens

### Test Case 3: Search with brackets in subject filter

**Given** an email with subject: `[nas-backup.example.com] Backup completed`

**Request**:
```json
{
  "using": ["urn:ietf:params:jmap:core", "urn:ietf:params:jmap:mail"],
  "methodCalls": [
    ["Email/query", {
      "accountId": "...",
      "filter": { "subject": "[nas-backup.example.com]" }
    }, "0"]
  ]
}
```

**Expected**: Email is returned

**Actual**: No results returned

### Test Case 4: Simple token search (works correctly)

**Request**:
```json
{
  "using": ["urn:ietf:params:jmap:core", "urn:ietf:params:jmap:mail"],
  "methodCalls": [
    ["Email/query", {
      "accountId": "...",
      "filter": { "subject": "backup" }
    }, "0"]
  ]
}
```

**Result**: Works correctly - returns all emails with "backup" in subject

## Analysis

The behavior suggests that the search engine:

1. **Tokenizes** the filter string on special characters (`.`, `-`, `[`, `]`, etc.)
2. **Performs AND matching** on all resulting tokens
3. **Does not preserve** the original string for substring matching

For example, `nas-backup.example.com` is likely tokenized into:
- `nas`
- `backup`
- `example`
- `com`

And the query returns emails containing ALL these tokens anywhere in the subject, rather than emails containing the literal substring.

## Impact

- Users cannot search for domain names or hostnames in subjects (e.g., `server.example.com`)
- Users cannot search for hyphenated terms as exact phrases
- Users cannot search for bracketed prefixes common in automated notifications (e.g., `[JIRA-1234]`, `[GitHub]`, `[server.example.com]`)
- Notification emails from systems like Synology NAS, Jenkins, or monitoring tools (which typically use `[hostname]` or `[system-name]` prefixes) cannot be filtered accurately

## Suggested Fix

The `subject` filter should perform literal substring matching (case-insensitive) on the raw subject string, without tokenization. This aligns with RFC 8621's description and user expectations.

Alternatively, if full-text search tokenization is intentional for performance reasons, consider:
1. Adding a separate `subjectContains` filter for exact substring matching
2. Documenting this behavior as a deviation from the RFC
3. Supporting quoted strings for exact phrase matching (e.g., `"nas-backup.example.com"`)

## Workaround

Currently, users must search for individual tokens (e.g., `backup` instead of `nas-backup.example.com`) and manually filter results client-side, which is impractical for automated filtering scenarios.

guimard:
## Summary

Alternative approach to #2131 for fixing JMAP Email/query subject and address filter substring matching.

**Problem:** The `uax_url_email` tokenizer (used by `KEEP_MAIL_AND_URL` analyzer) splits on `-`, `[`, `]`, `.` etc., making it impossible to match substrings like `nas-backup`, `example.com`, or `alice-test@domain.tld`.

**This PR's approach:** Replace the analyzer itself rather than switching to WildcardQuery.

- Introduces a `substring_analyzer` using `whitespace` tokenizer + `word_delimiter_graph` filter + `asciifolding` + `lowercase`
- `word_delimiter_graph` splits tokens at non-alphanumeric boundaries (e.g. `nas-backup.example.com` → `nas`, `backup`, `example`, `com`)
- Standard `MatchQuery` with `Operator.And` then naturally matches substrings without needing WildcardQuery

### Comparison with PR #2131 (WildcardQuery approach)

| Aspect | #2131 (WildcardQuery) | This PR (substring_analyzer) |
|--------|----------------------|------------------------------|
| Query type | WildcardQuery with `*term*` | Standard MatchQuery with AND |
| Performance | Slower (wildcard scan) | Faster (inverted index lookup) |
| Analyzer change | None | New analyzer on subject + address fields |
| Reindexation | Not required | **Required** |
| Scope | Subject only | Subject + address fields (fixes MAILBOX-401 too) |

### Trade-offs

- **Pro:** Better query performance (inverted index vs wildcard scan)
- **Pro:** Fixes both subject AND address hyphen matching (MAILBOX-401)
- **Pro:** Cleaner approach — no WildcardQuery wrapping needed
- **Con:** Requires reindexation of existing data
- **Con:** Changes the analyzer, which affects how all data is tokenized

## Changes

- **TmailMailboxMappingFactory**: Add `substring_analyzer`, apply to subject, from/to/cc/bcc name and address fields
- **TmailCriterionConverter**: Override `manageAddressFields` with `Operator.And` on name/address clauses to prevent false positives
- **Tests**: Re-enable 3 MAILBOX-401 hyphen address tests, add 3 new subject substring tests

## Test plan

- [ ] CI passes all 161 TmailOpenSearchIntegrationTest tests
- [ ] Subject search matches substrings with hyphens (`nas-backup`)
- [ ] Subject search matches substrings with dots (`example.com`)
- [ ] Address search matches hyphenated local parts (`alice-test`)
- [ ] Address search matches hyphenated domains (`domain-test.tld`)
- [ ] Existing address exact match still works
- [ ] No false positives on address search

Closes #2130

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email address field searching to match across name, address, domain, and raw address variations.

* **New Features**
  * Introduced substring-based search analyzer for enhanced matching across email fields.

* **Tests**
  * Re-enabled address matching tests and added new test coverage for searches with special characters, hyphens, and compound names.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tmail-backend/mailbox/opensearch/src/main/java/com/linagora/tmail/mailbox/opensearch/TmailCriterionConverter.java
+++ b/tmail-backend/mailbox/opensearch/src/main/java/com/linagora/tmail/mailbox/opensearch/TmailCriterionConverter.java
@@ -30,6 +30,8 @@
 
 package com.linagora.tmail.mailbox.opensearch;
 
+import static org.apache.james.backends.opensearch.IndexCreationFactory.RAW;
+
 import jakarta.inject.Inject;
 
 import org.apache.james.mailbox.model.SearchQuery;
@@ -120,7 +122,27 @@ public class TmailCriterionConverter extends DefaultCriterionConverter {
                 .build()
                 .toQuery();
         }
-        return super.manageAddressFields(headerName, value);
+        String fieldName = getFieldNameFromHeaderName(headerName);
+        return new BoolQuery.Builder()
+            .should(new MatchQuery.Builder()
+                .field(fieldName + "." + JsonMessageConstants.EMailer.NAME)
+                .query(new FieldValue.Builder().stringValue(value).build())
+                .operator(Operator.And)
+                .build().toQuery())
+            .should(new MatchQuery.Builder()
+                .field(fieldName + "." + JsonMessageConstants.EMailer.ADDRESS)
+                .query(new FieldValue.Builder().stringValue(value).build())
+                .operator(Operator.And)
+                .build().toQuery())
+            .should(new MatchQuery.Builder()
+                .field(fieldName + "." + JsonMessageConstants.EMailer.DOMAIN)
+                .query(new FieldValue.Builder().stringValue(value).build())
+                .build().toQuery())
+            .should(new MatchQuery.Builder()
+                .field(fieldName + "." + JsonMessageConstants.EMailer.ADDRESS + "." + RAW)
+                .query(new FieldValue.Builder().stringValue(value).build())
+                .build().toQuery())
+            .build().toQuery();
     }
 
     @Override
--- a/tmail-backend/mailbox/opensearch/src/main/java/com/linagora/tmail/mailbox/opensearch/TmailMailboxMappingFactory.java
+++ b/tmail-backend/mailbox/opensearch/src/main/java/com/linagora/tmail/mailbox/opensearch/TmailMailboxMappingFactory.java
@@ -73,6 +73,7 @@ public class TmailMailboxMappingFactory implements MailboxMappingFactory {
     private static final String NGRAM = "ngram";
     private static final String NGRAM_ANALYZER = "ngram_analyzer";
     private static final String FILENAME_ANALYZER = "filename_analyzer";
+    private static final String SUBSTRING_ANALYZER = "substring_analyzer";
     private static final String NGRAM_TOKENIZER = "ngram_tokenizer";
     private static final String WHITESPACE_TOKENIZER = "whitespace";
     private static final String LOWERCASE = "lowercase";
@@ -166,7 +167,7 @@ public class TmailMailboxMappingFactory implements MailboxMappingFactory {
                     .properties(ImmutableMap.of(
                         JsonMessageConstants.EMailer.NAME, new Property.Builder()
                             .text(new TextProperty.Builder()
-                                .analyzer(KEEP_MAIL_AND_URL)
+                                .analyzer(SUBSTRING_ANALYZER)
                                 .fields(RAW, new Property.Builder()
                                     .keyword(new KeywordProperty.Builder().normalizer(CASE_INSENSITIVE).build())
                                     .build())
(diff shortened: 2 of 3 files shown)


chibenwa (changes requested):
Important limitations

1. No real linguistic tokenization

With WHITESPACE_TOKENIZER, you get no real language-aware processing:

no stemming (manger ≠ mange)

no stopwords (le, de, et)

no proper punctuation handling


For “human” text (descriptions, articles, comments), this is weak.

2. WORD_DELIMITER_GRAPH can be too aggressive

It may generate many tokens that are sometimes useless:

"iPhone12ProMax"
→ i, phone, 12, pro, max

This is not always desirable.

--------

My understanding:

- great proposal for address fields! We may consider this as we saw it solves some bugs
- but it's an absolute no-go for subject that needs more analysis than exact match

PR #2149, opened 2026-02-09
[FIX] DownloadAll was leaking files

chibenwa:
## Summary by CodeRabbit

* **Refactor**
  * Improved attachment retrieval with added leak-tracking for better resource safety.
  * Reworked download flow to a staged process and added a parsing-based fallback when direct blob extraction yields nothing, increasing reliability for edge-case attachments.


Arsnael (commented):
Overall ok but don't you need this to be merged first on james side though? https://github.com/apache/james-project/pull/2937

PR #2202, opened 2026-02-25
Add canUpgrade/isPaying to domain subscription message

rezk2ll:
## Problem

The domain subscription handler currently handles DNS validation, storage quota, and rate limiting when a domain subscription message arrives from RabbitMQ. However, there is no way to set `canUpgrade` and `isPaying` at the domain level.

These two flags are read per-user by `SaaSCapabilityFactory` (exposed via JMAP capabilities) and the `IsPaying` mailet matcher. Unlike rate limiting, which has a built-in domain-level fallback in the mailets, `canUpgrade`/`isPaying` have no domain fallback — they are only read from the user table. This means there is currently no way to control these values for all users of a domain at once.

## Solution

Add `canUpgrade` and `isPaying` as optional fields to `SaaSDomainValidSubscriptionMessage`. When a domain subscription message arrives with these fields:

1. **Store domain-level defaults** as new columns (`can_upgrade`, `is_paying`) in the existing `domains` table. This follows the same pattern used for domain-level rate limiting — no new table is created.

2. **Propagate to all existing users** in that domain by iterating through `UsersRepository.listUsersOfADomainReactive` and upserting each user's `SaaSAccount` in the `user` table.

3. **Provision future users automatically** via a new `SaaSAccountProvisionListener` that listens for INBOX creation events. When a new user's INBOX is created, the listener reads the domain defaults and applies them to the new user.

Both fields are `Optional<Boolean>` for backward compatibility — existing messages without these fields are processed exactly as before. When only one field is provided, the other uses the default value from `SaaSAccount.DEFAULT`.

On domain cancellation, the domain-level defaults are cleared (columns set to null).

### DIFF
--- a/tmail-backend/apps/distributed/src/main/java/com/linagora/tmail/james/app/DistributedSaaSModule.java
+++ b/tmail-backend/apps/distributed/src/main/java/com/linagora/tmail/james/app/DistributedSaaSModule.java
@@ -24,6 +24,7 @@ import java.util.function.Function;
 
 import jakarta.inject.Named;
 
+import org.apache.james.events.EventListener;
 import org.apache.james.user.api.UsernameChangeTaskStep;
 
 import com.datastax.oss.driver.api.querybuilder.schema.CreateTable;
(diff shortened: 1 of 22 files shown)


chibenwa:
@rezk2ll thanks for the contrib!

@quantranhong1999 can we add a github ticket for this? And take over needed evolutions so that we can define those option domain wide and not need to define those on a user basis.

quantranhong1999:
> the canUpgrade is meant to prevent the users from attempting or trying to change the plan ( because they technically can't )

Regarding the business logic, can a domain's plan be upgraded? Or there is only 1 plan for B2B?

Arsnael:
Dumb question though (or not):

If any newly  created member of a saas b2b domain is being created, as they should all have the same pattern (aka isPaying = true and canUpgrade = false), why don't you just send a user creation payload with those fields, simply enough?

I dont think the values could be something else? canUpgrade => no need to show a message for that to end user, not concerning for him
isPaying => well if admin stops paying technically they all get disabled in the ldap ?

Arsnael:
Let's discuss that in tomorrow's meeting :)

PR #2226, opened 2026-03-03
ISSUE-2214 Fix Saas domain activation with rate limiting

Arsnael:
Honestly I feel like we doing here some ugly hack on domains... Not fond of it (opened to suggestions if can get it done better)

Have two remaining things I think:
- cassandra tests: where do I put those? Feels wrong to put them in saas-cassandra or saas-rabbitmq modules (as it would mean injecting the other module)... creating an other module under saas just for cassandra rabbitmq tests maybe?
- cache with TmailCassandraDomainList: does not seem to be working (cf disabled tests on latest commit). It doesn't seem we use cache in our environments anyways... do we need it (in which case I will try to fix it) or should we just disable it for TmailCassandraDomainList?

Also I think postgres we can just create an other ticket and tackle it later, this PR is big enough already and postgres fix ain't priority right now  

## Summary by CodeRabbit

* **New Features**
  * Domains now have an "activated" flag; domain list and rate-limiting flows respect activation state and a new Cassandra-backed domain list implementation is available.
  * Rate limiting API accepts an optional activation flag to record activation when configuring limits.

* **Documentation**
  * Added upgrade instructions to add the activated column to the domains table.

* **Tests**
  * Expanded integration and unit tests, introduced reusable contract-style tests and probes to cover activation and rate-limiting scenarios.

### DIFF
--- a/tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/JmapSaasContract.scala
+++ b/tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/JmapSaasContract.scala
@@ -21,7 +21,7 @@ package com.linagora.tmail.james.common
 import java.util.Optional
 
 import com.linagora.tmail.common.probe.SaaSProbe
-import com.linagora.tmail.james.common.JmapSaasContract.{DOMAIN_SUBSCRIPTION_ROUTING_KEY, QUOTA_ROOT, SUBSCRIPTION_ROUTING_KEY, TEST_ACCOUNT_ID, TEST_DOMAIN, TEST_PASSWORD, TEST_USER}
+import com.linagora.tmail.james.common.JmapSaasContract.{DOMAIN_SUBSCRIPTION_ROUTING_KEY, QUOTA_ROOT, RATE_LIMITATION, SUBSCRIPTION_ROUTING_KEY, TEST_ACCOUNT_ID, TEST_DOMAIN, TEST_PASSWORD, TEST_USER}
 import com.linagora.tmail.james.common.probe.{DomainProbe, RateLimitingProbe}
 import com.linagora.tmail.rate.limiter.api.model.RateLimitingDefinition
 import com.linagora.tmail.saas.model.SaaSAccount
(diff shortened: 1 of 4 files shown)


tmail-backend/data/data-cassandra/src/test/java/com/linagora/tmail/domainlist/cassandra/TmailCacheDomainListTest.java
```diff
@@ -77,6 +78,7 @@ void containsShouldBeCached(CassandraCluster cassandra) throws DomainListExcepti
             .hasSize(1);
     }
 
+    @Disabled("ISSUE-2214: issue with TmailCassandraDomainList and cache")
```
chibenwa: Can we try to fix that?

Arsnael:
> We remain the PG part, correct?

The answer was part of the initial description: 
"Also I think postgres we can just create an other ticket and tackle it later, this PR is big enough already and postgres fix ain't priority right now"

Follow up PR ok then if we can spare the time I dont mind

Arsnael:
Ready for review

tmail-backend/integration-tests/jmap/jmap-integration-tests-common/src/main/scala/com/linagora/tmail/james/common/JmapSaasContract.scala
```diff
@@ -322,6 +323,70 @@ trait JmapSaasContract {
...
+
+    awaitAtMostTenSeconds.untilAsserted(() => assertThat(server.getProbe(classOf[RateLimitingProbe]).getRateLimiting(domain))
+      .isEqualTo(RateLimitingDefinition.builder()
+        .mailsSentPerMinute(10)
+        .mailsSentPerHours(100)
+        .mailsSentPerDays(1000)
+        .mailsReceivedPerMinute(20)
+        .mailsReceivedPerHours(200)
+        .mailsReceivedPerDays(2000)
+        .build()))
```
quantranhong1999: This assertion can be merged into the above test `domainShouldBeUnactivatedWhenDomainSubscriptionSetUpRateLimiting`, to save 1 test running time?

tmail-backend/data/data-cassandra/src/test/java/com/linagora/tmail/domainlist/cassandra/TmailCassandraDomainListTest.java
```diff
@@ -0,0 +1,121 @@
...
+    void emptyActivatedShouldReturnDomain(CassandraCluster cassandra) {
+        CqlSession testingSession = cassandra.getConf();
+
+        testingSession.execute(String.format("INSERT INTO domains (domain) VALUES ('%s')", DOMAIN_1.asString()));
+
+        SoftAssertions.assertSoftly(softly -> {
+            softly.assertThat(Throwing.supplier(() -> domainList().containsDomain(DOMAIN_1)).get()).isTrue();
+            softly.assertThat(Throwing.supplier(() -> domainList().getDomains()).get()).containsOnly(DOMAIN_1, Domain.LOCALHOST /*default domain*/);
+        });
+    }
```
chibenwa: I think we should not return the domain here

We likely should require the operator to:
 - ALTER the table
 - Set existing domains to "activated"
 - and then and only then upgrade james.

chibenwa (changes requested):

tmail-backend/data/data-cassandra/src/test/java/com/linagora/tmail/domainlist/cassandra/TmailCassandraDomainListTest.java
Arsnael: cf => https://github.com/linagora/tmail-backend/pull/2230 (last two commits)

please read the description as I have a few concerns that need to be clarified if possible :)

PR #2230, opened 2026-03-05
ISSUE-2214 Fix Cassandra Saas domain activation with rate limiting (activated null = not activated)

Arsnael:
Variant of https://github.com/linagora/tmail-backend/pull/2226

Look at last two commits for the variant: activated null = unactivated

I also added something I missed yesterday with domain rate limit routes:
- can add rate limits on an unexisting domain (activated = false)
- still return not found if ratelimit exist but domain not activated (wasn't 100% sure here, let me know if you want a different behavior)

Regarding the refactoring: still have two methods for set rate limiting in RateLimitingRepository (with and without activated boolean). I think we can reduce to one.

1. Can keep the original if it's ok to have domain rows with activated null in case rate limit set before domain, treated as unactivated by TmailCasandraDomainList

OR 

2. can always set and override activated with the correct value (but adds unnecessary column overrides, and pre-reads to know if domain exists or not)

I prefer 1, but want to be sure.

## Release Notes

* **New Features**
  * Introduced domain activation capability with an "activated" field to manage domain status in the system.
  * Added rate limiting probes and repositories to support rate limiting configuration per domain.
  * Implemented new domain subscription handling with contract-based testing infrastructure for improved reliability.

* **Tests**
  * Established contract-based test patterns for domain subscription and rate limiting scenarios.

* **Chores**
  * Updated database schema to include the activated field in the domains table.

### DIFF
--- a/docs/modules/ROOT/pages/tmail-backend/webadmin.adoc
+++ b/docs/modules/ROOT/pages/tmail-backend/webadmin.adoc
@@ -649,14 +649,13 @@ curl -XPUT http://ip:port/domains/{domain}/ratelimits
 }'
 ....
 
-Notes:
-- null values are allowed (means no limit set)
-- can add rate limits to a domain before its activation
+Note: null values are allowed (means no limit set)
 
 Return codes:
 
 - `204` The domain rate limits updated successfully
 - `400` Invalid request
+- `404` Domain does not exist
 
 === Get rate limits of a domain
 ....
@@ -682,7 +681,7 @@ Return codes:
 
 - `200` get the rate limits of the domain successfully
 - `400` Invalid request
-- `404` Domain does not exist or is not activated
+- `404` Domain does not exist
 
 == Domain contacts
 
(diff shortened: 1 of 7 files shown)


tmail-backend/integration-tests/jmap/memory-jmap-integration-tests/src/test/java/com/linagora/tmail/james/MemoryJmapSaaSTest.java
```diff
@@ -120,4 +120,34 @@ public void quotaForUserShouldBeUserByDefaultWhenDomainAndUserQuotaDefined() {
     public void domainShouldBeRemovedWhenDomainSubscriptionDisabled() {
 
     }
+
+    @Disabled("Memory app does not support RabbitMQ consumer")
```
chibenwa: Is it worth it then to have MemoryJmapSaaSTest ?

Can't we rather just drop the test class if it tests nothing?

quantranhong1999 (approved):

PR #2243, opened 2026-03-10
JAMES-4189 Ability to list users by domain

chibenwa:
Depends on https://github.com/apache/james-project/pull/2973

## Release Notes

* **New Features**
  * The quota management system now supports retrieving and managing users at the domain level, enabling more granular quota tracking across different domains.


quantranhong1999 (commented):
Read it.

PR #2245, opened 2026-03-10
ISSUE-2159 Webadmin route to manage labels

chibenwa:
## Summary by CodeRabbit

* **New Features**
  * Added webadmin API endpoints to create, list, update, and delete user labels with support for display name, color, description, and keyword.

* **Documentation**
  * Added detailed documentation for the JMAP label endpoints including examples, payloads, and status codes.

* **Tests**
  * Added end-to-end API tests covering label lifecycle, validation, error cases, and responses.

### DIFF
--- a/tmail-backend/webadmin/webadmin-jmap-labels/src/main/java/com/linagora/tmail/webadmin/label/LabelRoutes.java
+++ b/tmail-backend/webadmin/webadmin-jmap-labels/src/main/java/com/linagora/tmail/webadmin/label/LabelRoutes.java
@@ -1,0 +1,175 @@
+/********************************************************************
+ *  As a subpart of Twake Mail, this file is edited by Linagora.    *
+ *                                                                  *
+ *  https://twake-mail.com/                                         *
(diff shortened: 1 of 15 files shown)


tmail-backend/webadmin/webadmin-jmap-labels/src/main/java/com/linagora/tmail/webadmin/label/LabelRoutes.java
```diff
@@ -0,0 +1,284 @@
...
+            throw ErrorResponder.builder()
+                .statusCode(HttpStatus.BAD_REQUEST_400)
+                .type(ErrorResponder.ErrorType.INVALID_ARGUMENT)
+                .message("'%s' is required and must not be empty", field)
+                .haltError();
+        }
+        return body.get(field).asText();
+    }
+
+    private Option<Color> optionalColor(JsonNode body, String field) {
```
Arsnael: The all class seems using scala conversions  around Option instead of using Optional directly... why? Scala has not much to do here

PR #2248, opened 2026-03-11
ISSUE-2212 Implement KeywordEmailQueryViewOptimizer

Arsnael:
Need the listener actually to be able to add proper integration tests for the optimizer

WIP draft for now

## Summary by CodeRabbit

* **New Features**
  * Added keyword-based email query filtering for improved search performance and resilience.

* **Configuration**
  * Introduced `view.keyword.query.enabled` option (default: disabled) to control keyword filtering behavior.

* **Documentation**
  * Updated configuration documentation with the new keyword query setting.


tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/method/KeywordEmailQueryViewOptimizer.scala
```diff
@@ -0,0 +1,99 @@
...
+  private def matchesInKeywordSortedByReceivedAt(request: EmailQueryRequest): Boolean =
+    configuration.isEmailQueryViewEnabled &&
+      request.filter.nonEmpty &&
+      keywordOnlyFilter(request) &&
+      request.sort.contains(Set(Comparator.RECEIVED_AT_DESC))
+
+  private def keywordOnlyFilter(request: EmailQueryRequest): Boolean =
+    request.filter.get match {
+      case condition: FilterCondition =>
+        condition.hasKeyword.nonEmpty &&
```
quantranhong1999: IMO, we should also filter the input keyword that matches only the concerned keywords: flagged (important flag) + user flags.

We do not index the keyword view for non-concerned keywords like `$seen`. If `Email/query` with `hasKeyword: $seen` then likely this would return empty (which may not be correct).

quantranhong1999:
> Need the listener actually to be able to add proper integration tests for the optimizer

Yes, it is. Hopefully I can push the listener work this afternoon.

tmail-backend/jmap/extensions/src/main/scala/com/linagora/tmail/james/jmap/TMailJMAPModule.scala
```diff
@@ -22,11 +22,14 @@ import java.io.FileNotFoundException
...
 import com.github.fge.lambdas.Throwing
 import com.google.common.collect.ImmutableSet
+import com.google.inject.multibindings.Multibinder
 import com.google.inject.name.Named
 import com.google.inject.{AbstractModule, Provides, Singleton}
+import com.linagora.tmail.james.jmap.method.KeywordEmailQueryViewOptimizer
 import com.linagora.tmail.james.jmap.oidc.JMAPOidcConfiguration
 import com.linagora.tmail.james.jmap.settings.ReadOnlyPropertyProvider
 import org.apache.commons.configuration2.{Configuration, PropertiesConfiguration}
+import org.apache.james.jmap.method.EmailQueryOptimizer
```
chibenwa: Question: do we have dedicated JMA tests covering keywords?

quantranhong1999:
> Yes, it is. Hopefully I can push the listener work this afternoon.

FYI, it is more or less ready at https://github.com/linagora/tmail-backend/pull/2250.

Arsnael:
Rebased on top of https://github.com/linagora/tmail-backend/pull/2250 and reworked the code.

I would say ready for review (but not for merge, so will stay on draft until prior necessary work is merged)

tmail-backend/apps/memory/src/main/java/com/linagora/tmail/james/app/modules/jmap/MemoryKeywordEmailQueryViewListenerModule.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.linagora.tmail.james.app.modules.jmap;
+
+import org.apache.james.events.EventListener;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.multibindings.Multibinder;
+import com.linagora.tmail.james.jmap.event.KeywordEmailQueryViewListener;
+
+public class MemoryKeywordEmailQueryViewListenerModule extends AbstractModule {
```
quantranhong1999: IMO, we should not bind either the keyword view listener or keyword view optimizer for Memory app. Otherwise, we may have the keyword view cleanup issue for E2E Label test cf https://github.com/linagora/tmail-backend/pull/2250#discussion_r2916527850.

`KeywordEmailQueryMethodContract` for Cassandra/Postgres app is enough IMO.

chibenwa (commented):
Ok with me so far.

Let's remove the memory part IMO

PR #2252, opened 2026-03-12
ISSUE-2251 Handle b2b saas user created case

Arsnael:
So quickly to explain some design choices (feel free to discuss):
- I consider that a b2b saas user should have canUpgrade=false (we dont show banner) but should also have isPaying to true (we dont want it to send messages via the free gateway but the premium one correct?)
- i deserialize two types of messages to treat them as two distinct cases (a bit like with domain). I hesitated with just put other fields of the user message sub optional except internalEmail and canUpgrade, but was afraid of a potential mess

Feel free to disagree and comment :)

## Summary by CodeRabbit

* **New Features**
  * Support for distinct user-created events (B2B) and subscription updates, with handling that preserves paying status while honoring upgrade capability.
* **Tests**
  * Added integration and unit tests validating parsing, routing and handling of user-created and subscription messages, including canUpgrade=false scenarios.
* **Documentation**
  * Added optional RabbitMQ settings for user-created exchange and routing key.


quantranhong1999:
> Underlying PG and Cassandra implems are to use dedicated DAOs operating atop can_upgrade and is_paying to be added columns of the domains table.
getSaaSAccount(Username username) returns the user value, then fallback to the domain value then defaults

So we go back to do the alternative proposed a while ago: 1 domain subscription message published by the SaaS team, no longer 1 message for 1 user subscription of the domain?

PR #2316, opened 2026-04-10
Update James SHA-1

Arsnael:
Still failing on the postgres server app tests

This https://github.com/apache/james-project/pull/2996 is breaking the tests for me.

It does not auto adjust space storage, it seems the broker service is trying to claim 100gb disk space. Our CI has 82GB available? (I have less locally)

It breaks for me on james too, but maybe the tests were green because the apache ci is beefy..?

Need to think of a fix on james side

Thinking jvm properties or conf option in activemq properties conf file?

## Summary by CodeRabbit

* **Refactor**
  * Modernized event serialization/deserialization across the system: serializers now return structured success/failure results instead of throwing, improving error handling and consistency.
* **Tests**
  * Updated tests to unwrap and assert serializer result payloads via new accessors.
* **Chores**
  * Updated submodule reference to a newer commit.


Arsnael:
Fix proposal => https://github.com/apache/james-project/pull/2999

PR #2321, opened 2026-04-13
Fix master build

quantranhong1999:
- adapt event serializer refactoring
- adapt blob store DAO API changes

## Summary by CodeRabbit

* **Refactor**
  * Improved blob storage operations with type-safe data handling patterns.
  * Enhanced event serialization infrastructure with result-based error handling across the event bus system.
  * Consolidated internal APIs for event processing to improve system reliability and type safety.

### DIFF
--- a/tmail-backend/blob/blobid-list/src/main/java/com/linagora/tmail/blob/blobid/list/SingleSaveBlobStoreDAO.java
+++ b/tmail-backend/blob/blobid-list/src/main/java/com/linagora/tmail/blob/blobid/list/SingleSaveBlobStoreDAO.java
@@ -18,7 +18,6 @@
 
 package com.linagora.tmail.blob.blobid.list;
 
-import java.io.InputStream;
 import java.util.Collection;
 
 import org.apache.james.blob.api.BlobId;
@@ -28,8 +27,6 @@ import org.apache.james.blob.api.ObjectNotFoundException;
 import org.apache.james.blob.api.ObjectStoreException;
 import org.apache.james.blob.api.ObjectStoreIOException;
 import org.reactivestreams.Publisher;
-
-import com.google.common.io.ByteSource;
 
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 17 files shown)


tmail-backend/blob/blobid-list/src/main/java/com/linagora/tmail/blob/blobid/list/SingleSaveBlobStoreDAO.java
```diff
@@ -29,8 +28,6 @@
 import org.apache.james.blob.api.ObjectStoreIOException;
 import org.reactivestreams.Publisher;
 
-import com.google.common.io.ByteSource;
-
 import reactor.core.publisher.Mono;
 
 public class SingleSaveBlobStoreDAO implements BlobStoreDAO {
```
quantranhong1999, line 33: I would eventually implement `MetadataAwareBlobStoreDAOContract` for `SingleSaveBlobStoreDAO` and `SecondaryBlobStoreDAO`. Not possible today because the S3 and Cassandra storage are not merged, but tested locally, these should be good.

tmail-backend/guice/distributed/src/main/java/com/linagora/tmail/event/TmailEventSerializer.java
```diff
@@ -163,12 +166,12 @@ private EventDTO toDTO(Event event) {
                 e.updatedLabel().readOnly());
             case LabelDestroyed e -> new LabelDestroyedDTO(e.getEventId().getId().toString(), e.username().asString(),
                 e.labelId().toKeyword());
-            default -> throw new IllegalStateException("Unexpected value: " + event);
-        };
+            default -> null;
```
chibenwa, line 169: ? Shouldn't we return empty ?

quantranhong1999: I cherry-picked from https://github.com/linagora/tmail-backend/pull/2316. Maybe @Arsnael has better insight about this?

Arsnael: The switch is wrapped into an Optional.ofNullable... Optional.ofNullable(null) == Optional.empty()

I thought it was better than wrapping each return in every switch case in an optional. But it's discutable I know

Arsnael (approved):

chibenwa (approved):
Let's move on, if that's localized bad code I can survive.

tmail-backend/guice/distributed/src/main/java/com/linagora/tmail/event/TmailEventSerializer.java
quantranhong1999: It is ok, I think.

PR #2385, opened 2026-05-08
ISSUE-2383 Support signature templates with combined user repository

### LINKED ISSUE #2383: DomainSignatureTemplateRoutesModule should be bound when combined user repository used
cf concern https://github.com/linagora/tmail-backend/pull/2376#discussion_r3205814696

We have `GET /domains/domainXYZ/signature-templates can not be found` on sandbox, which is using a combined user repository.

Indeed, given the current module chooser, we only bind `DomainSignatureTemplateRoutesModule` if we use the read-only LDAP user repository.

We should bind it upon combined user repository too.

chibenwa:


quantranhong1999 (approved):

Arsnael (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-backend-pr:2385

PR #2393, opened 2026-05-12
[BUILD] Run Tmail JMAP Postgres integration tests in the smoke testing mode

quantranhong1999:
Currently: 18:31  Integration Tests :: JMAP :: Postgres

Previously, we ran the whole PostgreSQL test suite as part of the ongoing Postgres development process. Apply similarly smoke testing, like distributed or Postgres JMAP ITs on James.

### DIFF
--- a/tmail-backend/integration-tests/jmap/postgres-jmap-integration-tests/pom.xml
+++ b/tmail-backend/integration-tests/jmap/postgres-jmap-integration-tests/pom.xml
@@ -143,6 +143,11 @@
                     <reuseForks>true</reuseForks>
                     <forkCount>2</forkCount>
                     <argLine>-Xms1024m -Xmx2048m</argLine>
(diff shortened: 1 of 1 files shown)


Arsnael (approved):

chibenwa:
Are we sure?

I see value in running all those tests, I would prefer not to take this shortcut. At least without trying seriously other options and ideas to reduce build time.

quantranhong1999:
> I see value in running all those tests, I would prefer not to take this shortcut. At least without trying seriously other options and ideas to reduce build time.

Fair. I think adopting the user isolation strategy would eventually make JMAP postgres tests a lot faster.

chibenwa:
Yes that is the idea 

I recall we explicitly said we won't run all it tests on distributed but at least we will do on pg

Closing for now

Arsnael:
Docker image published for this PR: linagora/tmail-backend-pr:2393

PR #2415, opened 2026-05-26
refactor(imap-delegation-ext): use new method names in James

felixauringer:
This adapts the IMAP Auth Delegation extension to the changes in the IMAP authentication logic introduced in https://github.com/apache/james-project/pull/3044.
It also changes the [documentation](https://github.com/linagora/tmail-backend/blob/master/docs/modules/ROOT/pages/tmail-backend/imap-extensions/imapAuthDelegationExtension.adoc) to match the actual code and be consistent between AUTH PLAIN and LOGIN.

The following changes for the IMAP LOGIN command:

- Now honors if plain authentication is disabled.
- Use authentication (user that authenticated) / authorization (user whose identity is assumed by the authenticated user) in variable names to be closer to the docs / the SASL RFC.

The following changes for AUTH PLAIN:

- Use authentication (user that authenticated) / authorization (user whose identity is assumed by the authenticated user) in variable names to be closer to the docs / the SASL RFC.

EDIT: The following is outdated.

I generally find it confusing that the correct syntax (according to the documentation) is `authentication_id+authorize_id` for LOGIN and `authorize-id+authenticate-id` for AUTH PLAIN.
If that is something that can still change, I would advise to use the syntax `<authorization-id>+<authentication-id>`. The order would match the one used by SASL PLAIN (RFC 4616). Additionally, the `+` character is already used for plus email addressing (or subaddressing), so that might collide.

### DIFF
--- a/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TMailAuthenticateProcessor.java
+++ b/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TMailAuthenticateProcessor.java
@@ -23,7 +23,6 @@ import java.util.List;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.james.core.Username;
-import org.apache.james.imap.api.display.HumanReadableText;
 import org.apache.james.imap.api.message.request.ImapRequest;
 import org.apache.james.imap.api.message.response.StatusResponseFactory;
 import org.apache.james.imap.api.process.ImapSession;
@@ -48,14 +47,13 @@ public class TMailAuthenticateProcessor extends AuthenticateProcessor {
     }
 
     @Override
-    protected void doPlainAuth(String initialClientResponse, ImapSession session, ImapRequest request, Responder responder) {
+    protected void parseAndDoPlainAuth(String initialClientResponse, ImapSession session, ImapRequest request, Responder responder) {
         AuthenticationAttempt authenticationAttempt = parseDelegationAttempt(initialClientResponse);
         if (authenticationAttempt.isDelegation()) {
-            doAuthWithDelegation(authenticationAttempt, session, request, responder);
+            doPasswordAuthWithDelegation(authenticationAttempt, session, request, responder);
         } else {
-            doAuth(authenticationAttempt, session, request, responder, HumanReadableText.AUTHENTICATION_FAILED);
+            doPasswordAuth(authenticationAttempt, session, request, responder);
         }
-        session.stopDetectingCommandInjection();
     }
 
     private AuthenticationAttempt parseDelegationAttempt(String initialClientResponse) {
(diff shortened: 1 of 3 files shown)


chibenwa:
Cool @felixauringer for preparing this PR ^^

I really appreciate, but also needs the "this effort is indeed not needed to merge things on James side". Sometime james changes breaks twake mail and it's fine for LINAGORA as a company to invest a bit of tim to fix things.

Cheers,

Benoit

chibenwa:
> Additionally, the + character is already used for plus email addressing (or subaddressing), so that might collide.

Subadressing is not to be used in auth, it will not collide. We needed a safe char and as it might be used by end users we needed something easy and "presentable" to show.

This convenience is niche, scoped to Twake mail, we are fine...

tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TMailAuthenticateProcessor.java
```diff
@@ -80,10 +78,10 @@ private AuthenticationAttempt handleTwoTokens(List<String> tokens) {
...
-        String authId = StringUtils.substringAfter(localPart, DELEGATION_SPLIT_CHARACTER);
-        String delegateId = StringUtils.substringBefore(localPart, DELEGATION_SPLIT_CHARACTER);
-        Username authUser = Username.of(authId).withDefaultDomain(user.getDomainPart());
-        Username delegateUser = Username.of(delegateId).withDefaultDomain(user.getDomainPart());
-        return delegation(authUser, delegateUser, tokens.get(1));
+        String authorizationId = StringUtils.substringBefore(localPart, DELEGATION_SPLIT_CHARACTER);
+        String authenticationId = StringUtils.substringAfter(localPart, DELEGATION_SPLIT_CHARACTER);
+        Username authenticationUser = Username.of(authenticationId).withDefaultDomain(user.getDomainPart());
+        Username authorizationUser = Username.of(authorizationId).withDefaultDomain(user.getDomainPart());
+        return delegation(authorizationUser, authenticationUser, tokens.get(1));
```
chibenwa:
```suggestion
        String authorizationId = StringUtils.substringAfter(localPart, DELEGATION_SPLIT_CHARACTER);
        String authenticationId = StringUtils.substringBefore(localPart, DELEGATION_SPLIT_CHARACTER);
        Username authenticationUser = Username.of(authenticationId).withDefaultDomain(user.getDomainPart());
        Username authorizationUser = Username.of(authorizationId).withDefaultDomain(user.getDomainPart());
        return delegation(authorizationUser, authenticationUser, tokens.get(1));
```

Let's retain the same semantic than before

Mimicing subaddressing getting `authuser+targetuser@domain` feels more natural...

felixauringer: I have adapted the documentation and the code to match this.

Before this PR, the docs for AUTH PLAIN were different.

quantranhong1999:
Hello @felixauringer, I could not push the James module update to your branch. Therefore, I opened another PR: https://github.com/linagora/tmail-backend/pull/2418.

Thanks!

felixauringer:
> Hello @felixauringer, I could not push the James module update to your branch. Therefore, I opened another PR: #2418.
> 
> Thanks!

Thanks, then I'll close this PR.

PR #2433, opened 2026-06-08
[ENHANCEMENT] CalDavCollect: Take RRT into account to check if the us…

chibenwa:
…er is concerned

## Summary by CodeRabbit

* **Refactor**
  * Consolidated iCalendar recipient-matching into a single injected component to simplify and unify processing across calendar sync flows.
* **Bug Fixes**
  * Improved handling of attendee aliases so invites sent to an alias are correctly recognized and synchronized to the invitee’s calendar.
* **Tests / Chores**
  * Added integration coverage for alias delivery and a build declaration to support the change.

### DIFF
--- a/tmail-backend/tmail-third-party/openpaas/pom.xml
+++ b/tmail-backend/tmail-third-party/openpaas/pom.xml
@@ -61,6 +61,10 @@
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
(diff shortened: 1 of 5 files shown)


chibenwa:
Seen with @guimard 

We do need to allow sabre to itip events that do not concerns the user,

Arsnael:
Docker image published for this PR: linagora/tmail-backend-pr:2433

PR #2435, opened 2026-06-09
[ENHANCEMENT] CalDavCollect: Take RRT into account to check if the us…

chibenwa:
…er is concerned

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified CalDAV synchronization: removed recipient/organizer gating so non-REPLY calendar messages are always forwarded for sync; REPLY messages still skip sync when the recipient is explicitly listed as an attendee.

* **Tests**
  * Updated integration tests to match the revised synchronization behavior, removing cases for conditional recipient filtering.


chibenwa:
Cc @vttranlina

Arsnael:
Docker image published for this PR: linagora/tmail-backend-pr:2435

quantranhong1999 (approved):

Arsnael (approved):

PR #2457, opened 2026-06-29
ISSUE-2456 Close TMail AMQP consumers upon shutdown

### LINKED ISSUE #2456: Make sure Tmail AMQP consumers close correctly upon TMail shutdown
cf the same issue on TCalendar: https://github.com/linagora/twake-calendar-side-service/pull/847

```
During the investigation of flaky CI tests, we found that several AMQP consumers implement Closeable, but their close() methods are not called when the application or Docker container is stopped/restarted.

The current lifecycle only invokes methods registered with Guice @PreDestroy. Because these consumers were missing @PreDestroy, their RabbitMQ subscriptions were not explicitly disposed during shutdown.

This can lead to non-graceful shutdown behavior, and in test/CI scenarios may leave consumers active longer than expected, causing messages to be consumed by stale consumers and making tests flaky.
```

Let's check if Tmail's AMQP consumers have the issue, if yes, fix them.

chibenwa:
Closes #2456

Several TMail AMQP consumers implement `Closeable` and dispose their RabbitMQ subscription in `close()`, but `close()` was **not** annotated `@PreDestroy`. James only invokes cleanup on `@PreDestroy`-annotated methods (tracked by the onami `PreDestroyModule` and disposed in `GuiceJamesServer.stop()`), so these `close()` methods were never called on shutdown.

Consequences (same root cause as twake-calendar-side-service#847):
- non-graceful shutdown on `docker stop`/restart, RabbitMQ subscriptions not disposed;
- in tests/CI, stale consumers keep consuming messages, causing flakiness.

### Fix

Annotate `close()` with `@PreDestroy` (mirrors the existing `RabbitMQAndRedisEventBus.stop()` precedent) on:

1. `TWPSettingsConsumer`
2. `SaaSSubscriptionConsumer`
3. `SaaSDomainSubscriptionConsumer`
4. `RabbitMQEmailAddressContactSubscriber`
5. `SabreContactsConsumer`

The `close()` bodies already do the right thing (`Optional.ofNullable(disposable).ifPresent(Disposable::dispose)`); disposing the subscription terminates the `Flux.using(...)` which auto-invokes `Receiver::close`, so no further changes were needed.

Consumers 1-4 are bound as Guice `@Singleton`. `SabreContactsConsumer` was a just-in-time binding injected into the singleton `SabreContactsOperator`; it is now explicitly bound `in(Scopes.SINGLETON)` in `OpenPaasContactsConsumerModule` so `@PreDestroy` reliably fires.

Checkstyle and compilation of the affected modules pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown handling for several background services, helping them clean up more reliably when the app stops.
  * Reduced the risk of lingering message consumers or subscription processes after restart or container shutdown.
  * Strengthened initialization for a contact-related service so it is consistently available when the app runs.

### DIFF
--- a/tmail-backend/jmap/extensions-rabbitmq/src/main/java/com/linagora/tmail/james/jmap/RabbitMQEmailAddressContactSubscriber.java
+++ b/tmail-backend/jmap/extensions-rabbitmq/src/main/java/com/linagora/tmail/james/jmap/RabbitMQEmailAddressContactSubscriber.java
@@ -28,6 +28,7 @@ import java.io.Closeable;
 import java.nio.charset.StandardCharsets;
 import java.util.Optional;
 
+import jakarta.annotation.PreDestroy;
 import jakarta.inject.Inject;
 import jakarta.inject.Named;
 
@@ -140,6 +141,7 @@ public class RabbitMQEmailAddressContactSubscriber implements Startable, Closeab
             });
     }
 
+    @PreDestroy
     @Override
     public void close() {
         Optional.ofNullable(messageConsume).ifPresent(Disposable::dispose);
--- a/tmail-backend/saas/saas-rabbitmq/src/main/java/com/linagora/tmail/saas/rabbitmq/settings/TWPSettingsConsumer.java
+++ b/tmail-backend/saas/saas-rabbitmq/src/main/java/com/linagora/tmail/saas/rabbitmq/settings/TWPSettingsConsumer.java
@@ -25,6 +25,8 @@ import java.io.Closeable;
 import java.time.Duration;
 import java.util.Optional;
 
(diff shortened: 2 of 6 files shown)


chibenwa:
<details><summary>🤖 Session stats</summary>

19 turns | 146s | in=2584 out=7742 cache_read=213841 cache_write=27377 | cost=0.5880 USD

</details>

quantranhong1999 (approved):

chibenwa (commented):
+1

Arsnael (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-backend-pr:2457

PR #2487, opened 2026-07-07
[TMail] ISSUE-2405 Team mailbox migration via scoped IMAP login

### LINKED ISSUE #2405: Team mailbox migration
## Why ?

Need to be easily able to connect onto the team mailbox and migrate emails to it.

## What ?

Idea was:

```
a0 LOGIN admin+teammailbox@linagora.com adminPass
```

And then map the mailbox layout to the team mailbox.

Similarily members can do this too:

```
a0 LOGIN member+teammailbox@linagora.com memberPass
```

(Which is a new cool feature!)

## How ?

Play with `TmailPathConverter.Factory` and present the "team mailbox scoped" layout if a team mailbox is delegated. (`TeamMailboxScopedPathConcerter`)

We *may* need to have `PathConverter.Factory`  return a mono in order 

We will nee to ensure provisionning upon connect uses correct MailboxPatch - Likely by not supplying the path directly but going through `PathConverter.Factory` 

We will need to test IMAP commands used for migration: 
 - [ ] LOGIN
 - [ ] LIST
 - [ ] NAMESPACE
 - [ ] SELECT
 - [ ] FETCH
 - [ ] STORE
 - [ ] EXPUNGE
 - [ ] COPY
 - [ ] MOVE

chibenwa:
Closes #2405

## Why

Migrating emails into a team mailbox (or letting a member migrate their own mail into a shared team mailbox) is awkward today: off-the-shelf tools like `imapsync` copy folders relative to the account root and do not know about the `#TeamMailbox.` prefix. This PR lets a member/manager log in **scoped** to a single team mailbox, which is then presented as the root layout (`INBOX`, `Sent`, ...).

## What

Reusing the existing `+` delegation syntax:

```
a0 LOGIN member+teamMailbox@domain memberPass
```

When the authorization identifier (`teamMailbox`) designates an existing team mailbox the authenticated user is a member/manager of, the session is scoped to that team mailbox. `LIST`/`SELECT`/`STATUS`/`FETCH`/... then operate on the team subtree presented as the root layout, so a migration tool copies it folder-by-folder as if it were a personal account.

This implements the direction discussed in the issue thread (give members and managers the right to *impersonate* a team mailbox, membership resolved via `TeamMailboxRepository`).

## How

- **`TeamMailboxImpersonationResolver`** — given `(authenticationId, authorizationId)`, resolves the authorization id to a `TeamMailbox` and checks the authenticated user is a member/manager (via `TeamMailboxRepository.listTeamMailboxes(user)`). Returns empty otherwise, so regular user-to-user `+` delegation is preserved.
- **`TMailLoginProcessor` / `TMailAuthenticateProcessor`** — on a `+` login that resolves to a team mailbox, the member is authenticated as themselves (no user delegation) and the resolved `TeamMailbox` scope is stashed on the `MailboxSession`.
- **`TeamMailboxScope`** — carries the scope in the `MailboxSession` attributes so both `PathConverter.Factory#forSession` overloads (`ImapSession` and `MailboxSession`) read it synchronously on every command, resolved once at login (no `Mono`-returning factory change).
- **`TeamMailboxScopedPathConverter`** — presents the scoped team mailbox as the root layout: `buildFullPath` rewrites relative names into the `#TeamMailbox` subtree, `mailboxName` echoes team paths back as root-relative names, `mailboxQuery` scopes `LIST` to the team subtree. Absolute `#TeamMailbox.` and `#user.` paths keep the regular behaviour.

The session user stays the **authenticated member**, so ACL enforcement keeps working transparently for `SELECT`/`FETCH`/`STORE`/`COPY`/`MOVE`/`EXPUNGE` (a plain member has no `DeleteMailbox` right, so folder `DELETE` may be denied — expected). Default provisioning is untouched: it runs on the member personal `INBOX` before the scope is applied, so no stray team folders are created.

## IMAP command checklist

- [x] LOGIN / AUTHENTICATE PLAIN (both `+` forms, incl. 3-token SASL)
- [x] LIST (scoped to the team subtree, relative names)
- [x] SELECT / STATUS / FETCH (via `buildFullPath`)
- [x] STORE / EXPUNGE / COPY / MOVE — enforced by ACL, path mapping via the scoped converter

## Tests

- `TeamMailboxScopedPathConverterTest` — unit coverage of `buildFullPath` / `mailboxName` / factory selection.
- `IMAPTeamMailboxMigrationIntegrationTest` — end-to-end scoped `LOGIN`, `LIST` (root layout), `SELECT`+`FETCH`, `STATUS`, and a non-member negative case.

## Notes / follow-ups

- Disambiguation: if a user delegation target and a team mailbox share the same local part, team-mailbox scoping wins when the user is a member.
- Admin bypass: this PR authorizes via **membership** (members and managers). A server admin who is *not* a member of the team mailbox is not yet granted scoped access; that can be added as a follow-up (e.g. via a `TMailAuthorizator`), mirroring the user-delegation admin semantics.
- I could not run the Maven build/tests in this sandboxed environment (Maven invocation is gated); please rely on CI to validate compilation and the new tests.

---
*Generated automatically*

## Summary by CodeRabbit

* **New Features**
  * IMAP SASL login now supports optional team-mailbox scoping for delegated access, exposing the scoped team mailbox subtree as the effective root.
  * Scoped sessions adjust namespace visibility and rewrite IMAP paths to ensure consistent browsing and mailbox navigation.
* **Bug Fixes**
  * Improved authorization and routing so scoped credentials only access the intended team mailbox (including correct admin/impersonation behavior).
  * Prevents namespace leakage (no unexpected `#TeamMailbox` prefixes in client-visible results).
* **Tests**
  * Added/expanded unit and integration coverage for scoped path conversion and end-to-end IMAP operations (LIST/SELECT/FETCH/COPY/MOVE/EXPUNGE), including admin and non-member negative cases.


chibenwa:
Addressed the CodeScene analysis on commit 93eba1a:

- **Bumpy Road Ahead** (critical gate, `TMailLoginProcessor.processRequest`): fixed. Extracted the delegation branch into a `processDelegatedRequest` guard-clause helper, so `processRequest` no longer carries nested conditional logic. Behavior is unchanged; the module compiles. Pushed as d6a6f08.

- **String Heavy Function Arguments** (non-critical, `TeamMailboxScopedPathConverter`): not actionable. The flagged string arguments (`mailboxName`, `finalReferencename`, `relative`) belong to `@Override` methods whose signatures are dictated by the `PathConverter` superclass contract — IMAP mailbox names are inherently strings there. Changing them would break the override and cannot be done in isolation, so I am leaving this one as-is.

CodeRabbit was still processing at the time of this pass; nothing else actionable had been posted.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

23 turns | 187s | in=4270 out=9456 cache_read=690043 cache_write=31898 | cost=0.9226 USD

</details>

chibenwa:
The only new activity is the CodeScene "String Heavy Function Arguments" biomarker on `TeamMailboxScopedPathConverter.java` (46.2% string arguments vs. a 39% threshold).

No code change is warranted here:

- The string-heavy methods (`buildFullPath(String mailboxName)`, `mailboxName(boolean, MailboxPath, MailboxSession)`, `mailboxQuery(String finalReferencename, String mailboxName, ImapSession)`) are `@Override`s whose signatures are dictated by the parent `PathConverter` / `TMailPathConverter` contract. They cannot be changed without breaking the override.
- The private helpers (`stripPrivateNamespace`, `scopedPath`, `referencesOtherNamespace`, ...) operate on IMAP mailbox names, which are strings by definition in the IMAP protocol.

Wrapping these inherent protocol strings in value types would add ceremony without improving safety, and would diverge from the surrounding `PathConverter` implementations. Leaving as-is.
---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

8 turns | 48s | in=3695 out=2604 cache_read=136425 cache_write=12498 | cost=0.2777 USD

</details>

chibenwa:
Addressed the CodeRabbit review in `e09783a6`:

**1. Scope team mailbox before provisioning (Major)** — Fixed. Previously the scope was applied *after* `doPasswordAuth`, so `authSuccess()` provisioned the personal `INBOX` before switching scope. The resolved scope is now *scheduled* at login time and applied by an overridden `provisionInbox()` right after `setMailboxSession(...)` and before provisioning, so a delegated migration login never touches the personal namespace. The pending scope is cleared in a `finally` around the auth call, so a failed scoped attempt can never leak onto a later login on the same connection. Applied symmetrically in both `TMailAuthenticateProcessor` and `TMailLoginProcessor`.

**2. Duplicate `scopeToTeamMailbox` helper (Nitpick)** — Fixed. The null-safe session scoping now lives in `TeamMailboxScope` (`scheduleScope` / `applyScheduledScope` / `clearScheduledScope`); the duplicated private helpers were removed from both processors.

**3. Missing command coverage (Nitpick)** — Added integration tests for `NAMESPACE` (asserting no `#TeamMailbox`/`#private` prefix leaks), `STORE`, `COPY`, `MOVE` and `EXPUNGE`, all exercised against the scoped session and verifying the operations target the team mailbox subtree.

Note: the local Maven toolchain was not available in this environment, so compilation/tests were verified by inspection against the James sources (signatures, checkstyle import order); CI will run the full build and the new integration tests.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

55 turns | 674s | in=4729 out=44888 cache_read=3502092 cache_write=83613 | cost=3.7340 USD

</details>

chibenwa:
The only new activity is the CodeScene delta gate-failure summary. It aggregates two findings, neither of which warrants a code change:

- **Code Duplication in `IMAPTeamMailboxMigrationIntegrationTest.java`** — this is a test file. The structural similarity between the `copy`/`expunge`/`move` scenarios is intentional: each test drives a distinct IMAP command against the team-mailbox subtree and stays self-contained and readable. CodeScene test-file duplication warnings are intentionally not acted upon here.

- **String Heavy Function Arguments in `TeamMailboxScopedPathConverter.java`** — this finding predates the last review round and is a borderline heuristic (46.2% strings vs a 39.0% threshold). The converter inherently maps IMAP mailbox names/paths, which are `String`-typed at this boundary, so the ratio reflects the domain rather than a design smell.

No behavioural or production-code change is needed.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

6 turns | 46s | in=3853 out=2788 cache_read=95262 cache_write=15898 | cost=0.2965 USD

</details>

chibenwa:
Overall good and agree though it clearly clashes with SASL work. CF https://github.com/linagora/tmail-backend/pull/2438)

I draft this and will undrafft (and rebase it) once SASL is merged.

Arsnael:
Docker images published for this PR: linagora/tmail-backend-pr:2487 linagora/tmail-migration-proxy-pr:2487

chibenwa (commented):
```
namespaceShouldNotLeakTeamMailboxPrefix com.linagora.tmail.imap.IMAPTeamMailboxMigrationIntegrationTest
[Failing for the past 1 build since #2](https://james-jenkins.lin-saas.com/job/Tmail%20build/view/change-requests/job/PR-2487/2/)
Took 0.84 sec
Error Details

Expecting actual:
  "* NAMESPACE (("" ".")) (("#user." ".")) (("#TeamMailbox." "."))
AAAB OK NAMESPACE completed.
"
not to contain:
  "#TeamMailbox
```

Rebase on linagora/master and fix this test

chibenwa:
Addressed the failing test; flagging the rebase because it is not a mechanical merge.

**`namespaceShouldNotLeakTeamMailboxPrefix` — fixed** (pushed `d93f45f2d`)

Made `TMailNamespaceSupplier` scope-aware: when the `MailboxSession` carries a team mailbox scope (`TeamMailboxScope.forSession(...)`), `otherUsersNamespaces` and `sharedNamespaces` return empty, so `NAMESPACE` reports `NIL` for them instead of leaking `#user` / `#TeamMailbox.`. Personal namespace (the team mailbox root) is unchanged.

Validated locally against the `linagora/tmail-backend` deps:
- `namespaceShouldNotLeakTeamMailboxPrefix` ✅
- `listShouldPresentTeamMailboxLayoutAsRoot` ✅ (regression guard)
- `imap-extensions` `compile` + `checkstyle:check` ✅

**Rebase on `linagora/master` — needs your call**

`git rebase linagora/master` hits the SASL clash you anticipated: JAMES-4210 (#2438) is now on master and it **deleted** `TMailLoginProcessor` / `TMailAuthenticateProcessor`, moving the `+` delegation parsing into `TMailPlainSaslMechanism` and wiring `TMailImapAuthPackage` to James' base `LoginProcessor` / `AuthenticateProcessor`. This PR establishes the team mailbox scope *by intercepting delegation inside those two processors* (`resolve(...)` → no-delegation `doPasswordAuth` → `TeamMailboxScope.scopeTo`), so it collides as a `modify/delete` conflict on both files.

This is a design decision rather than a conflict I can resolve blindly, and I didn't want to force an auth-path resolution that the integration test can't fully cover (the test drives `LOGIN`; if the base `AUTHENTICATE PLAIN` path now flows through the SASL mechanism, a naive re-introduction of the subclasses could leave `AUTHENTICATE PLAIN +teamMailbox` silently delegating to a non-existent user). Two ways forward:

1. **Re-home the scope resolution onto the new flow** — resolve the team mailbox scope where the SASL credentials are applied (so both `LOGIN` and `AUTHENTICATE PLAIN` are covered), and drop the two processor subclasses.
2. **Re-introduce thin `TMailLoginProcessor` / `TMailAuthenticateProcessor`** registered in `TMailImapAuthPackage`, keeping the interception approach but adapted to the base processors — only if `AUTHENTICATE` still routes through `parseAndDoPlainAuth`.

Happy to implement either once you confirm the direction (I'd lean (1) to stay aligned with the SASL work). The NAMESPACE fix above applies cleanly through whichever rebase you pick.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

61 turns | 765s | in=5344 out=40069 cache_read=3417569 cache_write=119760 | cost=3.9358 USD

</details>

chibenwa (commented):
> Re-home the scope resolution onto the new flow 

Please have a try at this!

Don't modify james-project.

If you hit a wall (likely IMO) reintroduce a minimal version of the 2 processors.

chibenwa:
Rebased on `linagora/master` and re-homed the scope resolution onto the new SASL flow. Force-pushed `f21253f01` (the 4 commits are squashed into one, because the JAMES-4210 `modify/delete` of the two processors makes a commit-by-commit replay meaningless).

## Did I hit the wall? Yes — but the reintroduced processors are genuinely minimal

Pure re-homing (SASL mechanism + factory only, no processors) is not reachable without touching `james-project`, for three compounding reasons in the new flow:

1. `TMailPlainSaslMechanism` (now in `sasl-extensions`) has no handle on the `ImapSession`/`MailboxSession`, so it cannot stash the scope anywhere the `PathConverter.Factory` can read it.
2. `JamesSaslAuthenticator.authenticatePassword` runs the delegation **authorization** check *before* returning `Success` — for `member+teamMailbox` it calls `authenticate(member).as(teamMailbox)`, the team mailbox is not a real user → `UNKNOWN_USER` → SASL `Failure`. We never even reach a success step.
3. Even past that, `AbstractAuthProcessor.handleSaslSuccess` builds the session via `authenticate(member).as(teamMailbox)` for any `authNId != authZId`, which fails identically.

So I took your fallback and reintroduced **thin** `TMailLoginProcessor` / `TMailAuthenticateProcessor` (≈25 lines each, no `processRequest` override, no `james-project` change). They sit on top of James' base processors and hook the new flow at exactly the two points above:

- **Authorization** — the constructor passes `jamesSaslAuthenticator.withExtraAuthorizator(teamMailboxImpersonationResolver.asAuthorizator())` to `super(...)`. The extra `Authorizator` returns `ALLOWED` when the authorization id is a team mailbox the member belongs to, `FORBIDDEN` otherwise. Given `AuthorizationState.combine`, this turns the team-mailbox case into `ALLOWED` while leaving every existing user-to-user delegation verdict (ALLOWED / FORBIDDEN / UNKNOWN_USER) untouched.
- **Session build** — an override of `handleSaslSuccess(...)` detects the team-mailbox identity via `resolveScope(...)` and, instead of delegating, builds a **non-delegated** session scoped to the team mailbox (`authenticate(member).withoutDelegation()` + `TeamMailboxScope.scopeTo`), then hands it to the inherited `doAuth(...)`.

The `+` **parsing itself stays entirely on the new flow** (`TMailPlainSaslMechanism`), so both `LOGIN` and `AUTHENTICATE PLAIN` are covered by construction — the processors add only the team-mailbox authorization + scoping decision.

## Fallout / simplifications

- `TeamMailboxScope` lost the `scheduleScope`/`applyScheduledScope`/`clearScheduledScope` dance and the `provisionInbox` override: because `handleSaslSuccess` now owns the session-supplier lambda, the scope is set on the `MailboxSession` before `authSuccess` provisions the INBOX, so provisioning already sees the team layout (no stray personal namespace) without a pending-attribute side-channel.
- `TMailImapAuthPackage` points back at `TMailLoginProcessor` / `TMailAuthenticateProcessor`.
- The `NAMESPACE` leak fix (`TMailNamespaceSupplier` scope-aware) and the scoped path converter are carried over unchanged.

## Build note

I could not run the Maven build locally: the sandbox `.m2` still holds a pre-JAMES-4210 `protocols-imap` (old 4-arg `LoginProcessor` constructor, no `handleSaslSuccess`) and the `james-project` submodule is empty, so nothing here — not even master's own `TMailIMAPModule` — compiles against it. The code is written against the pinned James SHA `2f7706bd` (`AbstractAuthProcessor` / `LoginProcessor` / `AuthenticateProcessor` / `JamesSaslAuthenticator` sources read directly at that commit). Please rely on CI for compilation + the integration tests.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

105 turns | 1411s | in=5553 out=87630 cache_read=10950234 cache_write=299458 | cost=10.6892 USD

</details>

tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TMailLoginProcessor.java
```diff
@@ -0,0 +1,73 @@
...
+    protected void handleSaslSuccess(SaslStep.Success success, ImapSession session, ImapRequest request, Responder responder, String successLog) {
+        SaslIdentity identity = success.identity();
+        Optional<TeamMailbox> teamMailboxScope = teamMailboxImpersonationResolver.resolveScope(identity);
+        if (teamMailboxScope.isPresent()) {
+            Username member = identity.authenticationId();
+            doAuth(() -> TeamMailboxScope.scopedSession(getMailboxManager(), member, teamMailboxScope.get()),
+                session, request, responder, member, identity.authorizationId(), successLog);
+            return;
+        }
+        super.handleSaslSuccess(success, session, request, responder, successLog);
```
chibenwa:
```suggestion
        if (teamMailboxScope.isPresent()) {
            Username member = identity.authenticationId();
            doAuth(() -> TeamMailboxScope.scopedSession(getMailboxManager(), member, teamMailboxScope.get()),
                session, request, responder, member, identity.authorizationId(), successLog);
        } else {
            super.handleSaslSuccess(success, session, request, responder, successLog);
        }
```

tmail-backend/integration-tests/imap/memory-imap-integration-tests/src/test/java/com/linagora/tmail/imap/IMAPTeamMailboxMigrationIntegrationTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+import com.google.inject.multibindings.Multibinder;
+import com.linagora.tmail.encrypted.MailboxManagerClassProbe;
+import com.linagora.tmail.james.app.MemoryConfiguration;
+import com.linagora.tmail.james.app.MemoryServer;
+import com.linagora.tmail.team.TeamMailbox;
+import com.linagora.tmail.team.TeamMailboxName;
+import com.linagora.tmail.team.TeamMailboxProbe;
+
+class IMAPTeamMailboxMigrationIntegrationTest {
```
chibenwa, line 52: If possible add a test with an IMAP admin: he should be able to impersonate any team mailbox

chibenwa:
Addressed the two review points in `f6f7e507d`:

**1. `handleSaslSuccess` early-return → if/else** (suggestion on `TMailLoginProcessor`)
Applied, and mirrored the same shape on `TMailAuthenticateProcessor` so both stay identical.

**2. Test with an IMAP admin impersonating any team mailbox**
This needed a small behaviour change first, then the test: previously `TeamMailboxImpersonationResolver` only authorized *members/managers*, and even if an admin got through, `TeamMailboxScope.scopedSession(...)` authenticated the session as the admin — who holds no ACL on the team mailbox, so `SELECT`/`FETCH` would fail.

Changes:
- `TeamMailboxImpersonationResolver` now injects `UsersRepository` and authorizes a login as a team mailbox when the authenticated user is a member/manager **or** a server administrator.
- `resolveScope` now returns a `TeamMailboxImpersonation(teamMailbox, sessionUser)`: a member acts as **himself** (keeping his restricted member/manager rights), while an administrator acts as the team mailbox **owner** (`team-mailbox@domain`, which owns the `#TeamMailbox` mailboxes), so ownership — not membership — grants full access.
- Added `administratorId` to the module test `usersrepository.xml` and two tests in `IMAPTeamMailboxMigrationIntegrationTest`: `adminShouldLoginWhenScopedToTeamMailboxHeIsNotMemberOf` and `adminShouldAccessTeamMailboxMessageWhenScoped` (admin is deliberately **not** a member). The existing `nonMemberShouldFailToLoginScopedToTeamMailbox` still guards non-admin non-members.

Note: Maven could not be run in my current sandbox (approval-gated), so I validated the changes by static review only — relying on CI to confirm compilation and the integration tests.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

72 turns | 1423s | in=4860 out=58816 cache_read=5075123 cache_write=113303 | cost=7.8212 USD

</details>

chibenwa:
The only new activity is CodeScene delta-analysis advisory flags (Quality Gate profile "The Bare Minimum"). None warrant a code change:

- **`handleSaslSuccess` — Excess Number of Function Arguments (5)** in both `TMailAuthenticateProcessor` and `TMailLoginProcessor`: this is an `@Override protected` method whose 5-argument signature is inherited verbatim from James' `AuthenticateProcessor`/`LoginProcessor`. It cannot be reduced without breaking the override contract.
- **`TMailAuthenticateProcessor` — Constructor Over-Injection (6)**: the constructor mirrors the parent processor's dependency-injection needs plus the single extra `TeamMailboxImpersonationResolver`; it is not gratuitous injection.
- **`TeamMailboxScopedPathConverter` — String Heavy Function Arguments**: mailbox names/paths are inherently `String`-based in the James IMAP path API this class bridges, so the ratio is intrinsic to the domain.
- **`IMAPTeamMailboxMigrationIntegrationTest` — Code Duplication**: this is a test file; per our conventions CodeScene test-file remarks are not acted upon (the three scenarios intentionally share structure to read consistently).

All four files remain above the code-health threshold enough that only the "bare minimum" gate flags them, and the impacts are minor (9.39/9.69). No change pushed.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

12 turns | 78s | in=3847 out=4781 cache_read=209965 cache_write=16729 | cost=0.4120 USD

</details>

chibenwa (commented):
Also add a test to ensure that it works with an admin configured at the IMAP only level

chibenwa:
Addressed the review: **added support and tests for an admin configured at the IMAP-only level.**

While writing the test I found that it did *not* work: `TeamMailboxImpersonationResolver` only recognized users repository administrators (`administratorId` in `usersrepository.xml`). An admin declared solely at the IMAP level (`auth.adminUsers.adminUser` in `imapserver.xml`) is unknown to the users repository, so the scoped login fell through to a (necessarily failing) user delegation onto the non-existent team mailbox user.

Changes:
- `TeamMailboxImpersonationResolver.resolveScope` now takes the IMAP admin users and honors them in addition to users repository administrators.
- `TMailLoginProcessor` / `TMailAuthenticateProcessor` capture `imapConfiguration.getAdminUsers()` via a `configure(ImapConfiguration)` override and pass them to `resolveScope`.
- `asAuthorizator()` deliberately ignores IMAP-level admins: James already grants them the delegation through its own `withAdminUsers` authorizator during `processRequest`, so only members/managers and users repository admins matter there.
- Test resource `imapserver.xml` now declares `imapadmin@domain.tld` as an IMAP-level admin (it is neither a users repository admin nor a team member), and two new tests assert that such an admin can log in scoped to the team mailbox and read its message: `imapConfiguredAdminShouldLoginWhenScopedToTeamMailboxHeIsNotMemberOf` and `imapConfiguredAdminShouldAccessTeamMailboxMessageWhenScoped`.

Note: my sandbox cannot compile the branch locally (the required `james-project` snapshot with the `JamesSaslAuthenticator`-based `LoginProcessor` constructor is not installed), so I validated statically and rely on CI for compilation and the integration tests.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

75 turns | 794s | in=5026 out=45818 cache_read=4338185 cache_write=126599 | cost=4.6066 USD

</details>

chibenwa:
The only new activity is from the CodeScene bot. None of the findings warrant a code change:

- **Constructor Over-Injection** on `TMailAuthenticateProcessor` / `TMailLoginProcessor` (6 args vs max 5): five of those parameters are simply forwarded verbatim to the James `super(...)` constructor (`AuthenticateProcessor` / `LoginProcessor`), and the sixth is the single `TeamMailboxImpersonationResolver` dependency this subclass adds. All are required Guice injections. Collapsing them into a parameter object would diverge from the upstream James processor signature these classes deliberately mirror, for no real benefit.
- **Excess Number of Function Arguments** on `handleSaslSuccess`: this is an `@Override` of the parent method — the signature is dictated by `AuthenticateProcessor`/`LoginProcessor` and cannot be changed here.
- **Code Duplication** in `IMAPTeamMailboxMigrationIntegrationTest`: this is a test file; the "duplication" is the intentional given/when/then structure shared across scenarios.

No changes pushed.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

10 turns | 98s | in=3841 out=3485 cache_read=130893 cache_write=9637 | cost=0.3668 USD

</details>

chibenwa:
No new actionable feedback since the last update — the only new activity is the CodeScene delta-analysis bot, whose findings were already addressed in the previous comment:

- **Constructor Over-Injection** on `TMailAuthenticateProcessor` / `TMailLoginProcessor` (6 args): five params are forwarded verbatim to the upstream James `super(...)` constructor these classes deliberately mirror, the sixth is the single `TeamMailboxImpersonationResolver` dependency this subclass adds. All required Guice injections — collapsing them would diverge from the upstream signature for no benefit.
- **Excess Number of Function Arguments** on `handleSaslSuccess`: an `@Override`, so the signature is fixed by the parent processor.
- **Code Duplication** in `IMAPTeamMailboxMigrationIntegrationTest`: test file, intentional given/when/then structure across scenarios.

No changes pushed.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

5 turns | 48s | in=3870 out=2499 cache_read=74979 cache_write=21774 | cost=0.3380 USD

</details>

chibenwa (commented):
+1

Arsnael:
Docker images published for this PR:
 - linagora/tmail-backend-pr:2487
 - linagora/tmail-backend-distributed-pr:ai-2487
 - linagora/tmail-migration-proxy-pr:2487

quantranhong1999:
Ci failures:

```
imapConfiguredAdminShouldAccessTeamMailboxMessageWhenScoped – com.linagora.tmail.imap.IMAPTeamMailboxMigrationIntegrationTest
1s
Error
Expecting actual:
  "AAAB OK LIST completed.
"
to contain:
  "* LIST (\HasNoChildren) "." "INBOX""
Stacktrace
java.lang.AssertionError: 
Expecting actual:
  "AAAB OK LIST completed.
"
to contain:
  "* LIST (\HasNoChildren) "." "INBOX"" 
	at com.linagora.tmail.imap.IMAPTeamMailboxMigrationIntegrationTest.imapConfiguredAdminShouldAccessTeamMailboxMessageWhenScoped(IMAPTeamMailboxMigrationIntegrationTest.java:253)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
```
and

```
adminShouldAccessTeamMailboxMessageWhenScoped – com.linagora.tmail.imap.IMAPTeamMailboxMigrationIntegrationTest
<1s
Error
Expecting actual:
  "AAAB OK LIST completed.
"
to contain:
  "* LIST (\HasNoChildren) "." "INBOX""
Stacktrace
java.lang.AssertionError: 
Expecting actual:
  "AAAB OK LIST completed.
"
to contain:
  "* LIST (\HasNoChildren) "." "INBOX"" 
	at com.linagora.tmail.imap.IMAPTeamMailboxMigrationIntegrationTest.adminShouldAccessTeamMailboxMessageWhenScoped(IMAPTeamMailboxMigrationIntegrationTest.java:231)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1604)
```

Seems related.

chibenwa:
Thanks — the two CI failures were a real bug in the admin impersonation commit, not flakiness. Fixed in a3576730a.

**Root cause.** `StoreMailboxManager::searchMailboxes` builds its result from two branches:

- `getBaseMailboxes` only runs when the query targets the *private* namespace (`MailboxQuery::isPrivateMailboxes`), so it never returns a `#TeamMailbox` mailbox;
- `getDelegatedMailboxes` returns the ACL-delegated ones, then filters out `mailbox.getUser().equals(session.getUser())`.

So a `#TeamMailbox` mailbox is only ever surfaced to a session that holds an ACL on it *and* does not own it. Members work because `addMember` grants them an ACL. But I was scoping the administrator session to `TeamMailbox::owner` (`team-mailbox@domain.tld`), which is precisely the path owner and holds no ACL — it fell out of both branches, hence `LIST` returning nothing while `NOOP`/login still succeeded. That is why only the two `LIST`-asserting admin tests failed.

**Fix.** The administrator session now runs as `TeamMailbox::self` (`marketing@domain.tld`) instead. `TeamMailboxRepository::createMailboxReliably` grants `FULL_RIGHTS` to `self` on every team mailbox folder at creation time, and `self` is not the path owner, so it survives the delegated-search filter and gets full access for `SELECT`/`FETCH`/`STORE`. It is also just the authorization id the admin typed, which makes the scoped session read naturally. Members keep running as themselves so their role-based ACL (member vs manager) still applies.

Note I could not run the build locally (the `james-project` submodule is not checked out in my working copy), so I am relying on CI to confirm the two tests now pass.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

47 turns | 610s | in=4426 out=41774 cache_read=2485819 cache_write=88239 | cost=3.1927 USD

</details>

chibenwa:
The only new activity since the last round is the CodeScene code-health gate. I reviewed each violation and none of them is actionable without changing upstream Apache James APIs, so no code change was made.

**`TMailAuthenticateProcessor` / `TMailLoginProcessor`**

- *Excess Number of Function Arguments — `handleSaslSuccess` has 5 arguments.* This method is an `@Override` of `AuthenticateProcessor.handleSaslSuccess` / `LoginProcessor.handleSaslSuccess` in `apache-james-imap-processor`. Its signature `(SaslStep.Success, ImapSession, ImapRequest, Responder, String)` is imposed by the parent class, and both implementations delegate back to `super.handleSaslSuccess(...)` with those same five arguments. Reducing the parameter count is not possible on our side.
- *Constructor Over-Injection — 6 arguments.* Five of those six (`MailboxManager`, `StatusResponseFactory`, `MetricFactory`, `PathConverter.Factory`, `JamesSaslAuthenticator`) are mandatory arguments of the `super(...)` constructor. Only `TeamMailboxImpersonationResolver` is added by TMail, and it is needed both to build the extra `Authorizator` passed to `super` and to resolve the team mailbox scope at authentication time. Bundling it with `JamesSaslAuthenticator` into a wrapper would shave off exactly one argument while adding an indirection, and the file would still fail the gate on the `handleSaslSuccess` rule above — so it buys nothing.

**`TeamMailboxScopedPathConverter`**

- *String Heavy Function Arguments (46.2% vs. 39.0% threshold).* The three public methods are overrides of `PathConverter`: `buildFullPath(String)`, `mailboxQuery(String, String, ImapSession)` and `mailboxName(boolean, MailboxPath, MailboxSession)`. IMAP hands mailbox names in as raw strings, and the private helpers (`referencesOtherNamespace`, `stripPrivateNamespace`, `scopedPath`, `teamName`) exist precisely to parse those strings before they become `MailboxPath` values. Introducing a value type at this boundary would only move the string parsing to the caller, which is the inherited interface.

**`IMAPTeamMailboxMigrationIntegrationTest`**

- *Code Duplication across 5 test methods.* This is a test file; the similarity is the shared IMAP connect/select/assert arrange-act-assert shape, which we deliberately keep explicit per test so a failure points straight at the scenario.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

20 turns | 116s | in=3987 out=7794 cache_read=407342 cache_write=25735 | cost=0.6768 USD

</details>

Arsnael:
Docker images published for this PR:
 - linagora/tmail-backend-pr:2487
 - linagora/tmail-backend-distributed-pr:ai-2487
 - linagora/tmail-migration-proxy-pr:2487

chibenwa:
@quantranhong1999 opinion ?

quantranhong1999:
> @quantranhong1999 opinion ?

I will have a look.

quantranhong1999:
Codex findings (testing the newly released GPT 5.6 Sol model):


1. **[High] Admin delegation is hijacked even when no team mailbox exists.**  
   [`teamMailbox()` only validates syntax](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TeamMailboxImpersonationResolver.java#L101-L120); the admin branch never calls `teamMailboxRepository.exists(...)`.

   Consequently:

   - `admin+typo@domain` can authenticate into a nonexistent scope.
   - More seriously, ordinary admin delegation such as `admin+alice@domain` is interpreted as team-mailbox impersonation whenever `alice` is a syntactically valid team name, even if `alice` is actually a normal user.

   This contradicts the claim that regular user delegation is preserved. Existence must be checked before choosing team-mailbox behavior.

2. **[High] Scoped sessions break IMAP’s case-insensitive `INBOX` rule.**  
   [`buildFullPath()` preserves the supplied casing](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TeamMailboxScopedPathConverter.java#L59-L68), unlike James’ normal converter, which canonicalizes `inbox`, `InBoX`, etc.

   Therefore `SELECT inbox`, `STATUS inbox`, or `COPY ... inbox` targets `marketing.inbox` rather than `marketing.INBOX`. This needs canonicalization in both direct path and LIST query handling.

3. **[Medium] LIST/LSUB reference-name composition is incorrect.**  
   [[Lines 93–96](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TeamMailboxScopedPathConverter.java#L93-L96)](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TeamMailboxScopedPathConverter.java#L93-L96) unconditionally append a delimiter after a non-empty reference.

   Examples:

   - `LIST "Sent" "*"` should match `Sent` and `Sent...`, but the query starts at `marketing.Sent.` and excludes `Sent`.
   - `LIST "Sent." "*"` becomes `marketing.Sent..*`.
   - The reference name is not Modified UTF-7 decoded, so non-ASCII references fail.

   IMAP concatenates reference and pattern verbatim; the converter should not invent a delimiter.

4. **[Medium/security] Admin identity disappears from audit records.**  
   The processor passes only `impersonation.sessionUser()` into [`[scopedSession](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TMailAuthenticateProcessor.java#L70-L77)`](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TMailAuthenticateProcessor.java#L70-L77), which then calls [`[authenticate(teamMailbox.self()).withoutDelegation()](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TeamMailboxScope.java#L60-L63)`](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TeamMailboxScope.java#L60-L63).

   For an admin, both `MailboxSession.user` and `loggedInUser` become `marketing@domain`; the actual authenticating admin is lost. At minimum, the IMAP AUTH audit entry will claim that the team mailbox authenticated itself. Preserve `authenticationId` as the logged-in/delegating identity.

5. **[Medium] Full-domain team-mailbox mode is bypassed.**  
   The [factory selects the scoped converter before checking `imap.teamMailbox.fullDomain.enabled`](https://github.com/linagora/tmail-backend/blob/a3576730ab25b061c5662cc826ce0a67ed503195/tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TMailPathConverterFactory.java#L38-L51). The scoped converter extends the normal converter, not `TMailFullDomainPathConverter`.

   Thus explicit paths such as `#TeamMailbox.marketing@domain.tld.INBOX` regress in full-domain deployments, despite the scoped converter promising that absolute paths retain regular behavior.

chibenwa:
Thanks @quantranhong1999 for running Codex over this — 4 of the 5 findings were real. Pushed 3 commits (`8310ed3ed`, `5b611a820`, `09b30575a`).

**1. [High] Admin delegation hijacked when no team mailbox exists — fixed.** Real, and the more serious half of the report is the right one: `admin+alice@domain.tld` was read as a team mailbox impersonation whenever `alice` parsed as a valid team name, landing the admin in an empty scope instead of `alice`'s mailbox. `sessionUser()` now gates on `teamMailboxRepository.exists(...)` before choosing the impersonation branch, so a non-existent team mailbox falls back to regular user-to-user delegation. Covered by `adminShouldKeepDelegatingOntoRegularUserWhoseNameParsesAsATeamMailboxName`.

Note on the `admin+typo@domain` sub-case: with the fix it no longer authenticates into a team mailbox scope, but the login still succeeds via James' own admin delegation, which allows an administrator to assume an unknown authorization id. That is upstream James behaviour (`withAdminUsers`), unchanged by and out of scope for this PR — I wrote a test asserting the login failed and it (correctly) did not, so I dropped it.

**2. [High] `INBOX` case sensitivity — fixed.** `buildFullPath` now applies the same canonicalization as James' default converter (IMAP-349). LIST/LSUB is deliberately *not* canonicalized: James does not do it in `mailboxQuery` either, and matching that is the point.

**3. [Medium] LIST/LSUB reference-name composition — fixed.** The unconditional trailing delimiter excluded the reference name itself (`LIST "Sent" "*"` missed `Sent`) and produced `marketing.Sent..*` for `"Sent."`. Removed: `PrefixedRegex` already skips the delimiter that follows its prefix, so passing the reference verbatim reproduces James' semantics exactly. The reference is now Modified UTF-7 decoded too.

**4. [Medium/security] Admin identity lost from audit records — fixed.** The session is now built as a delegation from the administrator onto the team mailbox, keeping him `loggedInUser` so the AUTH audit entry names him. Since no authorizator can grant a delegation onto a team mailbox (it is not a user of the users repository, `UserRepositoryAuthorizator` returns `UNKNOWN_USER`), the delegation is force-allowed at that one call site, `resolveScope` having already authorized the impersonation.

**5. [Medium] Full-domain mode bypassed — fixed.** `TeamMailboxScopedPathConverter` no longer extends `TMailPathConverter`; it delegates absolute `#TeamMailbox.` / `#user.` paths to the converter the session would have used had it not been scoped, so `imap.teamMailbox.fullDomain.enabled` is honoured. The factory picks the delegate.

**@coderabbitai nitpick on `isMember`** — valid, applied. `listMembers(teamMailbox)` is a single ACL lookup on one mailbox, versus `listTeamMailboxes(user)` which searched and filtered every team mailbox of the user.

Validation: `imap-extensions` 599 unit tests green (18 in `TeamMailboxScopedPathConverterTest`, up from 9), `IMAPTeamMailboxMigrationIntegrationTest` 17 green, checkstyle green.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

105 turns | 1304s | in=5801 out=51940 cache_read=8295015 cache_write=140337 | cost=6.8795 USD

</details>

chibenwa:
The only new feedback since the last round is the CodeScene delta analysis. I reviewed each of the four flagged files and none of them warrants a code change:

**`TMailAuthenticateProcessor.java` / `TMailLoginProcessor.java` — *Excess Number of Function Arguments* on `handleSaslSuccess`**

`handleSaslSuccess(SaslStep.Success, ImapSession, ImapRequest, Responder, String)` is an `@Override` of the method inherited from James' `AuthenticateProcessor`. Both implementations delegate back to `super.handleSaslSuccess(...)` on the non-impersonating path. The arity is imposed by the upstream contract; we cannot shrink it without diverging from James, and introducing a parameter object here would only move the arguments one level down while breaking the override.

**`TeamMailboxScopedPathConverter.java` — *String Heavy Function Arguments***

The three public methods (`buildFullPath`, `mailboxName`, `mailboxQuery`) are `@Override`s of the `PathConverter` interface, whose signatures are likewise fixed by James. The private helpers (`stripPrivateNamespace`, `sanitizeMailboxName`, `referencesOtherNamespace`) operate on the raw, still-unparsed IMAP mailbox names those overrides receive — a wrapper type there would be constructed and unwrapped within the same call, for no gain in type safety.

**`IMAPTeamMailboxMigrationIntegrationTest.java` — *Code Duplication***

Duplication in test code, where the repetition keeps each IMAP scenario readable on its own. Per our conventions we do not act on CodeScene findings in test files.

No commit pushed.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

10 turns | 52s | in=3844 out=3451 cache_read=153005 cache_write=17765 | cost=0.3608 USD

</details>

tmail-backend/imap-extensions/src/main/java/com/linagora/tmail/imap/TeamMailboxImpersonationResolver.java
```diff
@@ -107,11 +108,18 @@ private Optional<TeamMailbox> teamMailbox(Username authorizationId) {
...
+     * {@code alice} user rather than be read as a team mailbox impersonation.
+     *
      * Not {@link TeamMailbox#owner()}: James only ever surfaces {@code #TeamMailbox} mailboxes through
      * the ACL-delegated branch of its mailbox search, which discards the mailboxes the session user owns.
      * A session running as the owner would thus see an empty LIST.
      */
     private Optional<Username> sessionUser(Username authenticationId, TeamMailbox teamMailbox, Collection<String> imapAdminUsers) {
+        if (!exists(teamMailbox)) {
+            return Optional.empty();
+        }
```
chibenwa, line 122: I do not see a test case for this one. Is it missing?

chibenwa:
No code change needed for the latest activity.

The only new feedback since the last round is an auto-generated CodeRabbit reply acknowledging the previous fixes (existence check before team mailbox impersonation, `INBOX` canonicalization, LIST/LSUB reference composition, preserving the administrator identity, and the `isMember`/`listMembers` simplification) and marking its review threads as resolved. It contains no new review requests, no reviews were submitted, and no new inline comments were left.

---
*Generated automatically*

chibenwa:
<details><summary>🤖 Session stats</summary>

5 turns | 18s | in=3709 out=1114 cache_read=60448 cache_write=5973 | cost=0.1375 USD

</details>

quantranhong1999 (approved):

Arsnael:
Docker images published for this PR:
 - linagora/tmail-backend-pr:2487
 - linagora/tmail-backend-distributed-pr:ai-2487
 - linagora/tmail-migration-proxy-pr:2487

